PR #874, opened 2026-03-16
fix: production hardening — safe exports, dead imports, index guard

msaad00:
## Summary

- **Safe file exports**: All 12 export functions (JSON, SARIF, CycloneDX, SPDX, HTML, CSV, JUnit, Markdown, Prometheus, SVG, Graph, Badge) now wrapped with `_safe_export()` — catches `PermissionError`/`OSError` gracefully instead of crashing the scan
- **Dead import removed**: unused `hashlib` import left over from scan_id implementation
- **Index guard**: `server.packages[0].ecosystem` now falls back to `'unknown'` when package list is empty
- **Unused imports removed**: 5 dead imports from `output/__init__.py` (datetime, Path, uuid4, Any, Optional)

## Test plan

- [x] 115 scan/resolver/robustness tests pass
- [x] Pre-commit hooks pass (ruff, mypy, bandit)
- [x] Export error now shows clean message instead of traceback

### DIFF
--- a/src/agent_bom/parsers/compiled_parsers.py
+++ b/src/agent_bom/parsers/compiled_parsers.py
@@ -788,7 +788,7 @@ def parse_cargo_packages(directory: Path, *, resolve_versions: bool = False) -> 
     if cargo_lock.exists():
         current_name: Optional[str] = None
         current_version: Optional[str] = None
-        for raw_line in cargo_lock.read_text().splitlines():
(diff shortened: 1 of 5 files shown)


msaad00:
Superseded by #875 which includes all these fixes + OCI parser safety + test fix for Go ecosystem assertion

PR #1594, opened 2026-04-21
bump release version to 0.80.0

msaad00:
## Summary
- bump the repo-managed release version from `0.79.0` to `0.80.0`
- align release-managed docs, Docker/Helm metadata, registry manifests, and workflow pins with the new version
- unblock release publishing after the protected `v0.79.0` tag could not be moved

## Why
`v0.79.0` was already pushed before the release workflow permissions fix in `#1593`, and repository rules prevent retagging that protected tag. The clean path is to cut a new semver release from current `main`.

## Validation
- `python3 scripts/bump-version.py 0.80.0`
- `python3 scripts/bump-version.py 0.80.0 --check`
- commit hooks on commit (`check yaml/json/toml`, `ruff`, `ruff-format`, `mypy`, `bandit`)

### DIFF
--- a/.github/workflows/cve-freshness.yml
+++ b/.github/workflows/cve-freshness.yml
@@ -26,7 +26,7 @@ jobs:
         run: |
           uv run agent-bom sbom tests/fixtures/test-sbom.cdx.json -f sarif -o results.sarif || true
           if [ ! -f results.sarif ]; then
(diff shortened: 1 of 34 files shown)


agent-bom (approved):
Approving the release version bump so the fixed release workflow can ship as v0.80.0.

andres-linero (approved):

PR #1870, opened 2026-04-25
fix(ui): install lightningcss native binding in Docker

msaad00:
Summary

- install the Linux x64 GNU lightningcss native binding during the UI Docker dependency stage
- keeps host/package-lock behavior unchanged while making the Docker image build independent of npm optional-dependency omission

Validation

- git diff --check
- pre-commit hooks: whitespace, conflict, ruff/format/mypy/bandit skipped where no matching files

Could not run local `docker build` because the local Docker daemon is not running: `Cannot connect to the Docker daemon`.

Unblocks UI Validate failures like: `Cannot find module '../lightningcss.linux-x64-gnu.node'`.


msaad00:
Folded this Docker lightningcss native-binding fix into #1867 so the completeness telemetry PR carries the required UI Validate fix directly and we avoid a duplicate standalone PR.

PR #2002, opened 2026-04-27
chore: audit-2 P2 residue — RLS guard tests + retrigger race fix

msaad00:
## Summary

Closes the small actionable items from the audit-2 P2 list. Each item is either shipped with a real test/code fix or explicitly skipped with a rationale below — no scope creep.

### 1. Tenant RLS bypass guard now under test

`APIKeyMiddleware.dispatch` rejects any request that enters while an active `bypass_tenant_rls()` context is still set (\`src/agent_bom/api/middleware.py:805-810\`). That path is the defence-in-depth net for a \`with bypass_tenant_rls():\` block that somehow escapes its scope. It had no coverage. Adds two cases:

- \`test_api_key_middleware_rejects_request_when_tenant_rls_bypass_is_active\` — monkeypatches \`is_tenant_rls_bypassed()\` to \`True\` and asserts a 500 with \"Tenant isolation guard is not clean\".
- \`test_api_key_middleware_does_not_check_rls_bypass_when_postgres_disabled\` — proves the \`os.environ.get(\"AGENT_BOM_POSTGRES_URL\")\` short-circuit keeps the SQLite-only path off the postgres dependency.

### 2. \`scripts/retrigger_stranded_pr.sh\` race

Replaces the fixed \`sleep 2\` between \`gh pr close\` and \`gh pr reopen\` with a polling loop that watches \`repos/.../pulls/<n>\` until the PR state is observably \`closed\`. The previous form raced when GitHub had not propagated the close: the reopen call landed against a still-open PR and silently no-opped. Polls 10 × 1s then warns and falls through.

## Skipped (with rationale — not pulled into this PR)

- **\`ui/Dockerfile\` \`# pending-digest\` markers (3×)** — intentional per the comment block at the top of the file: \"for at most one day after a major bump.\" \`git blame\` shows they were added in #1983 (today, 2026-04-26), so still inside the documented window.
- **\`_ROLE_RULES\` no explicit \`/v1/posture/backpressure\` entry** — covered by \`(\"GET\", \"/v1/posture\", \"viewer\")\` via \`path.startswith\` in the rule loop. Explicit entry would be redundant.
- **\`_clean_graph_part\` empty-output fallback** — service and resource_type already fall back to \`\"unknown-service\"\` / \`\"resource\"\` literal placeholders. There is no source-of-truth alternative to chain to like provider has.
- **\`README:17\` cloud inventory** — the README header already reads \"agents, MCP servers, packages, containers, **cloud**, GPU, and runtime.\" Finding based on an older snapshot.

## Test plan

- [x] \`pytest tests/test_api_hardening.py -q\` — green.
- [x] \`bash -n scripts/retrigger_stranded_pr.sh\` — clean.
- [x] \`ruff check\` — clean.

No source-of-truth files moved; no public API change.

---

Follow-up to the user audit-2 P2 list (no single tracking issue).


msaad00:
auto-retrigger: required CI did not fire on head 97265b5a93892f832f82041a3db4c8a1a590678a (likely a GITHUB_TOKEN-authored 'Update branch' merge). Reopening to re-run checks.

msaad00:

agent-bom (approved):

andres-linero (approved):

PR #1996, opened 2026-04-27
feat(graph): add direct cloud_principal → agent MANAGES edge

msaad00:
## Summary

Closes audit P0 #1.

#1941 added \`principal → cloud_resource\` and \`cloud_resource → agent\` edges, but the principal was never wired directly to the agent. Single-hop "which principals can reach this agent?" queries had to traverse the intermediate cloud_resource node — fragile and silently degrades when callers iterate the immediate neighbours of an agent node.

This change adds the direct edge in \`_promote_cloud_origin_lineage\` right after the existing \`principal → resource MANAGES\` edge:

\`\`\`
principal_node_id  →  agent_id   (MANAGES, via=resource_node_id)
\`\`\`

The intermediate edges stay so the lineage is fully reconstructable; the new edge is purely additive. The \`via\` field on the evidence dict records that the relationship is mediated by a \`cloud_resource\` so consumers can distinguish direct ownership from cloud-mediated operation when they need to.

## Tests
- \`test_cloud_origin_metadata_becomes_lineage_nodes\` extended: asserts the direct \`principal → agent\` edge exists and \`evidence.via\` matches the cloud_resource id
- \`test_no_principal_agent_edge_when_principal_metadata_absent\` (new): guards the negative path — cloud_origin without cloud_principal must not emit the new edge or any service_account node

## Test plan
- [x] \`pytest tests/test_graph_builder.py -q\` — 34 passed (was 33)
- [x] \`ruff check\` — clean

PR #2008, opened 2026-04-27
chore: audit-4 P1/P2 hygiene bundle (5 fixes)

msaad00:
## Summary

Closes the verified-real findings from audit-4 in one focused PR. Each item is independently scoped, has a regression test or CI gate locking it in, and changes no public API.

### 1. Tenant quota races — atomic check + insert (P1)

\`enforce_*_quota\` did read-then-check, but the actual insert happened later in the route handler with **no lock**. Two concurrent requests both passed the check and both inserted, so a tenant exceeded its quota by N under load.

**Fix**: new \`tenant_quota_guard(tenant_id, *checks)\` context manager in \`agent_bom.api.tenant_quota\` that takes a per-tenant \`threading.Lock\`, runs the supplied check callables, and yields to the caller. Callers wrap the (check + insert) pair under the guard so the second concurrent request serialises on the lock and sees the first request's row in its check.

Updated four primary insert sites:
- \`api/routes/scan.py\` — submit-scan path (active + retained quotas).
- \`api/routes/schedules.py\` — create-schedule path.
- \`api/routes/observability.py\` — push-result path (retained quota).
- \`api/server.py\` scheduler-trigger path (active + retained quotas).

\`api/routes/fleet.py\` (sync path with 60+ line for-loop body) is the remaining race surface — admin-only, bigger refactor — tracked as a follow-up. Multi-replica clusters still need a Postgres advisory lock; that ships as a separate PR so the storage layer can carry the lock.

### 2. Missing UI relationship colors + Vitest invariant (P1)

\`RELATIONSHIP_COLOR_MAP\` in \`ui/lib/graph-schema.ts\` was missing six values that already exist in the Python \`RelationshipType\` enum: \`REMEDIATES\`, \`TRIGGERS\`, \`MANAGES\`, \`OWNS\`, \`PART_OF\`, \`MEMBER_OF\`. \`MANAGES\` was the freshest gap — #1996 added it for the new cloud_principal → agent edge, so those edges were rendering without color.

Added the six entries plus a new vitest invariant (\`tests/graph-schema-color-invariant.test.ts\`) that walks every \`RelationshipType\` value and fails when any is missing.

### 3. Workflow timeout-minutes sweep + CI gate (P1)

#1995 only swept \`release.yml\` + \`post-merge-self-scan\` + \`container-rescan\`. The other **29 jobs across 20 workflows** could still run unbounded.

Added \`timeout-minutes\` to all 29 missing jobs with sensible defaults (15m default; 20m security gates; 30m publish/build; 60m fuzzing batches). Added \`scripts/check_workflow_timeouts.py\` and wired it into the Security Scan job so any future workflow added without a timeout fails CI.

### 4. \`shutil.which()\` → absolute path TOCTOU (P2)

Two sites resolved a binary by bare name and let the subsequent \`subprocess\` resolve it again against PATH:
- \`proxy_sandbox.resolve_container_runtime\` — docker/podman pick.
- \`filesystem._scan_archive\` → \`_run_syft\` — syft fallback for tar archives.

Both now resolve via \`shutil.which\` once and pass the absolute path through.

### 5. \`O_NOFOLLOW\` on skill-bundle hash read (P2)

\`skill_bundles._sha256_file\` opened files via \`Path.open\` after a prior \`_ref_path_uses_symlink\` check — leaving a TOCTOU window where a symlink could be swapped between check and read. Now opens with \`os.O_RDONLY | os.O_NOFOLLOW\` so the kernel itself rejects a symlinked leaf.

## Test plan

- [x] \`pytest tests/test_api_tenant_isolation.py tests/test_proxy_sandbox.py tests/test_filesystem.py -q\` — **76 passed**.
- [x] \`scripts/check_workflow_timeouts.py\` — clean, every job has timeout-minutes.
- [x] \`ruff check\` on all touched files — clean.
- [x] \`mypy\` on touched modules — clean.
- [x] \`vitest run tests/graph-schema-color-invariant.test.ts\` — 2 cases pass.

## Issues addressed

Closes audit-4 P1/P2 hygiene findings shared in chat. No single GH issue tracks the bundle; each fix is referenced inline via \"audit-4 P1\" / \"audit-4 P2\" code comments.

### DIFF
--- a/.github/workflows/cflite-pr.yml
+++ b/.github/workflows/cflite-pr.yml
@@ -15,6 +15,7 @@ jobs:
   Build:
     if: github.event_name != 'pull_request'
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     permissions:
       contents: read
     steps:
@@ -28,6 +29,7 @@ jobs:
   PR-Fuzzing:
     if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     permissions:
       contents: read
       security-events: write
@@ -49,6 +51,7 @@ jobs:
   Batch-Fuzzing:
     if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-latest
+    timeout-minutes: 60
     needs: Build
     permissions:
       contents: read
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,6 +99,17 @@ jobs:
         # marker for at most one dependabot cycle (~24h).
         run: python scripts/check_docker_base_policy.py
 
+      - name: Workflow job timeout policy
+        # Closes the audit-4 P1 finding that #1995 only swept release.yml
+        # + post-merge-self-scan + container-rescan; the rest of
+        # .github/workflows/ still had jobs that could run unbounded and
+        # hold scarce runners. Every workflow job must declare
+        # timeout-minutes; reusable-workflow callers (`uses:`) inherit
+        # the timeout from the called workflow.
+        run: |
+          python -m pip install --quiet pyyaml
+          python scripts/check_workflow_timeouts.py
+
       - uses: ./.github/actions/setup-python
         with:
           python-version: '3.11'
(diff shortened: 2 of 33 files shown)


msaad00:
auto-retrigger: required CI did not fire on head f4323dd54e91f44ddb8783af1ef73b6a09830268 (likely a GITHUB_TOKEN-authored 'Update branch' merge). Reopening to re-run checks.

msaad00:

agent-bom (approved):

andres-linero (approved):

PR #2035, opened 2026-04-28
feat(helm): KEDA-driven autoscaling on control-plane API + published SLO

msaad00:
## Summary

Closes the highest-impact gap from the recent EKS scoring audit (**HPA custom metrics not wired** — composite 8.0/10 → projected 8.5/10 once this lands) by giving operators a first-class, queue-aware scaling path that targets bursty scanner workloads.

The chart's previous `autoscaling.enabled=true` shipped a CPU-only HPA. CPU lags the actual saturation signal on the workload pattern most agent-bom operators run (bursty scan jobs that build up backpressure long before CPU registers the load). KEDA fixes that by polling Prometheus for **leading-indicator metrics the API tier already exports**.

## What landed

- **`templates/controlplane-api-keda-scaledobject.yaml`** — KEDA `ScaledObject` with two default Prometheus triggers tuned to bursty fan-in:
  - rate-limit pressure (`rate(agent_bom_rate_limit_hits_total{bucket=~"global|tenant"}[1m])`)
  - API p99 latency (`histogram_quantile(0.99, ...)`)
  Operators can fully replace the trigger list via `keda.triggers` to plug in SQS / Kafka / queue-depth scalers.
- **`templates/controlplane-api-hpa.yaml`** — guard so the static CPU/memory HPA is suppressed when KEDA is on (KEDA generates its own; stacking two HPAs against the same Deployment fights itself).
- **`values.yaml`** — `controlPlane.api.autoscaling.keda` block (default `enabled: false`, no breaking change). Polling interval, cooldown, fallback-replicas on Prom outage, passthrough `advanced` for behavior policies.
- **`examples/eks-keda-values.yaml`** — production preset (min=3, max=12, 30s poll, 5min cooldown, scale-up burst policy, scale-down stabilization). Pairs with `eks-production-values.yaml`.
- **`site-docs/deployment/scaling-slo.md`** — published per-tier SLO with PromQL definitions, tuning guidance, and an **honest "what's still not measured" section** (no first-class `scan_jobs_active` gauge yet, no clustered Postgres benchmark yet — both flagged as the v0.83 follow-up that unblocks the "elastic at 1M edges" claim).

## Why this is the right scope

The audit identified two gaps blocking 8.5/10:
1. **HPA custom metrics not wired** ← addressed here.
2. **Clustered Postgres scale benchmark not published** ← explicitly called out in the SLO doc as the next missing piece, with `maxReplicas` guidance keyed off Postgres connection-pool capacity in the meantime.

Bundling those two would have grown this PR into infrastructure-dependent benchmark runs. The benchmark belongs in its own PR with a script + a captured run.

## Test plan
- [ ] CI Helm Profile Validate green (defaults — no KEDA emitted, static HPA emitted)
- [ ] CI Helm Profile Validate green for `examples/eks-keda-values.yaml` (KEDA emitted, static HPA suppressed)
- [ ] Manual `helm template -f eks-keda-values.yaml | grep -E "ScaledObject|HorizontalPodAutoscaler"` shows exactly one `ScaledObject` and zero static HPAs against `agent-bom-api`

### DIFF
--- a/deploy/helm/agent-bom/examples/eks-keda-values.yaml
+++ b/deploy/helm/agent-bom/examples/eks-keda-values.yaml
@@ -1,0 +1,74 @@
+# EKS preset — KEDA-driven autoscaling on the control-plane API tier.
+#
+# Use this preset when:
+#   * Scanner traffic is bursty (queue-depth / job-completion-rate spikes)
+#     and CPU/memory HPA lags the saturation signal.
+#   * You already run KEDA cluster-side (operator + CRDs installed).
+#   * Prometheus is reachable from the agent-bom namespace.
+#
+# Pair with eks-production-values.yaml for the full Postgres-backed control
+# plane, mTLS, ExternalSecrets, etc. — `helm install -f eks-production-values.yaml -f eks-keda-values.yaml`.
+#
+# What this preset does
+# ─────────────────────
+#   * Replaces the default CPU-only HPA on `agent-bom-api` with a KEDA
+#     ScaledObject driven by two Prometheus queries:
+#       - rate-limit pressure (saturation precursor)
+#       - API p99 latency (request-servicing health)
+#   * Floors at 3 replicas so a single-AZ blip cannot drop us under quorum.
+#   * Caps at 12 replicas to keep Postgres connection-pool usage bounded.
+#   * 30s poll, 5min cooldown — fast enough for scan-burst patterns,
+#     slow enough to avoid HPA flapping.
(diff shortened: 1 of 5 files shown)


msaad00:
auto-retrigger: required CI did not fire on head 1aab6497d9ea2b249ee79553452fb4bc2cc9b5a1 (likely a GITHUB_TOKEN-authored 'Update branch' merge). Reopening to re-run checks.

msaad00:

agent-bom (approved):

andres-linero (approved):

PR #2086, opened 2026-04-29
perf: parallelize AWS provider discovery

msaad00:
Summary

- run enabled AWS service discovery jobs through a bounded 4-worker executor instead of serial top-level service calls
- keep each worker on its own boto3 session and preserve deterministic result/warning aggregation order
- isolate Lambda and Step Functions parent warning buffers so parallel jobs do not share mutable warning state
- add regression coverage proving all enabled services run, ordered aggregation is preserved, and a slow first service does not block later services from starting

Why

AWS discovery covered the right services, but busy accounts could spend most of their scan time waiting on independent read-only APIs in sequence. This keeps the same read-only discovery behavior while reducing wall-clock latency for multi-service accounts.

Validation

- `pytest tests/test_cloud_aws.py -q` -> 11 passed
- `ruff check src/agent_bom/cloud/aws.py tests/test_cloud_aws.py`
- `mypy src/agent_bom/cloud/aws.py tests/test_cloud_aws.py`
- `git diff --check`

Refs #2085

### DIFF
--- a/src/agent_bom/cloud/aws.py
+++ b/src/agent_bom/cloud/aws.py
@@ -15,6 +15,8 @@ import json
 import logging
 import os
 import zipfile
+from collections.abc import Callable
+from concurrent.futures import ThreadPoolExecutor
 from email.parser import Parser as EmailParser
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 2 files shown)


msaad00:
auto-retrigger: required CI did not fire on head ebadf6d8b05e779578512b66553261a9f2b46c55 (likely a GITHUB_TOKEN-authored 'Update branch' merge). Reopening to re-run checks.

msaad00:

agent-bom (approved):

andres-linero (approved):

PR #2097, opened 2026-04-29
ci: stabilize dependabot lockfile normalization

msaad00:
## Summary
- serialize Dependabot UI lockfile normalization runs per PR
- push normalized lockfile commits with a strict force-with-lease against the PR head SHA
- keep the pull_request_target trust model: checkout base code only, copy only manifest files, and do not run PR-defined npm scripts

## Why
The normalizer currently commits from the trusted base checkout and then tries a regular push to the Dependabot branch. When the branch already contains the Dependabot commit or another normalization attempt, the push is non-fast-forward and the check fails even when the lockfile can be normalized safely.

## Validation
- parsed .github/workflows/dependabot-ui-lockfile-normalize.yml with PyYAML
- asserted HEAD_SHA is passed to the commit step and the force-with-lease refspec is present
- git diff --check

Refs #2096

### DIFF
--- a/.github/workflows/dependabot-ui-lockfile-normalize.yml
+++ b/.github/workflows/dependabot-ui-lockfile-normalize.yml
@@ -9,6 +9,10 @@ on:
 
 permissions: {}
 
+concurrency:
+  group: dependabot-ui-lockfile-normalize-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
 jobs:
   normalize-ui-lockfile:
     if: >-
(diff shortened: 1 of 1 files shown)


msaad00:
Status update: all CI checks for this workflow hardening PR are green. It is blocked only on code-owner review. This fixes the Dependabot UI lockfile normalize failure seen on #2096 by making the privileged pull_request_target normalizer push with a strict lease against the PR head SHA and serialize per-PR runs.

agent-bom (approved):
Approved. The workflow keeps the pull_request_target trust boundary and fixes the non-fast-forward normalizer push with per-PR concurrency plus a strict head-SHA lease.

agent-bom (approved):
Re-approved after rebase on current main. Scope remains the same: lease-protected Dependabot normalizer push and per-PR concurrency.

andres-linero (approved):

PR #2167, opened 2026-05-01
Fix Dependabot lockfile normalize loop

msaad00:
Summary
- compare the normalized UI lockfile against the PR-supplied lockfile instead of the trusted base branch index
- prevent the normalize workflow from committing a fresh no-op normalization commit on every reopen/synchronize

Why
#2164 kept moving to a new head SHA after every retrigger because the workflow always saw the expected Dependabot lockfile diff relative to main as a change to commit. That detached checks from the current PR head repeatedly.

Validation
- ruby -e 'require "yaml"; YAML.load_file(".github/workflows/dependabot-ui-lockfile-normalize.yml"); puts "workflow parses"'
- git diff --check

### DIFF
--- a/.github/workflows/dependabot-ui-lockfile-normalize.yml
+++ b/.github/workflows/dependabot-ui-lockfile-normalize.yml
@@ -67,6 +67,7 @@ jobs:
         # trusting the PR's `scripts` block.
         run: |
           cd ui
+          cp package-lock.json /tmp/ui-package-lock.from-pr.json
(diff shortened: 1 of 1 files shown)


agent-bom (approved):

agent-bom (approved):
Workflow fix is scoped to the normalize comparison and prevents repeat no-op force-pushes.

andres-linero (approved):

PR #2182, opened 2026-05-02
fix(version-utils): handle npm SemVer pre-release tags in version compare

msaad00:
## Summary
Closes the false-positive class where a package on a higher major than an advisory's fix bound was flagged as affected because the fix bound carried an npm-style pre-release suffix that PEP 440 cannot parse.

## Concrete repro

A real machine running `next@16.2.4` got flagged by `CVE-2023-46298` (Next.js missing cache-control header), whose affected range is `[0.9.9, 13.4.20-canary.13)` for `npm:next`.

Pre-fix behavior:
1. `packaging.version.Version("13.4.20-canary.13")` raises `InvalidVersion` — PEP 440 doesn't accept npm SemVer pre-release suffixes (`-tag.serial`).
2. `compare_version_order(16.2.4, 13.4.20-canary.13, npm)` caught the parse error and returned `None`.
3. The OSV/GHSA range matcher treats `None` as "unknown" and conservatively marks the package as affected (correct default for genuinely unknown comparisons, wrong here).

Result: `agent-bom check next@16.2.4 -e npm` reported `CVE-2023-46298` as affecting a Next.js 16 install, even though the fix landed in Next.js 13.

## Fix

Add a fall-back path inside `compare_version_order`: when the primary PEP 440 parse fails AND the input contains a recognised SemVer pre-release tag stem (`canary`/`beta`/`alpha`/`rc`/`pre`/`dev`/`nightly`/`next`/`snapshot`/`m`/`preview`), strip the suffix from both sides and retry the numeric `major.minor.patch` compare.

Technically lossy per SemVer 2.0 (`13.4.20-canary.X` is strictly a pre-release of `13.4.20`), but for OSV/GHSA introduced/fixed advisory bounds the `major.minor.patch` view is the correct safe answer: operators on a strictly higher `major.minor.patch` shouldn't be flagged.

## Regression coverage

- `tests/test_version_utils.py::test_npm_canary_prerelease_compare_avoids_false_positive` — direct repro of the next@16.2.4 FP plus the still-vulnerable cases (13.4.19, 1.0.0)
- `tests/test_version_utils.py::test_npm_other_semver_prerelease_tags_compare_correctly` — covers `rc`/`beta`/`alpha`/`pre`/`nightly` against higher stable releases

## Validation

- `ruff check` + `mypy` on touched files: passed
- `pytest tests/test_version_utils.py` — **44 / 44 pass** (42 existing + 2 new)
- pre-commit hooks (ruff, ruff-format, mypy, bandit): passed
- live: `agent-bom check next@16.2.4 -e npm` — no longer matches CVE-2023-46298 after this fix lands in the local DB lookup path

## Out of scope

- A separate latent bug in `normalize_version("1.0.0rc1", "pypi")` returns `1.0.0.postc1` (turns RC into POST-release) — this PR doesn't touch that path; existing tests for it pass; will open a follow-up.
- The "OSV record only has GIT range type events" issue (the upstream OSV mirror of CVE-2023-46298 has no semver ranges, only commit SHAs) — that's an upstream-data quality issue. agent-bom's local DB merges OSV + GHSA by CVE ID and ends up with the correct npm range from GHSA, so this fix is sufficient for the user-visible FP.

### DIFF
--- a/src/agent_bom/version_utils.py
+++ b/src/agent_bom/version_utils.py
@@ -429,7 +429,56 @@ def compare_version_order(left: str, right: str, ecosystem: str) -> int | None:
         right_norm = normalize_version(right, eco)
         return (Version(left_norm) > Version(right_norm)) - (Version(left_norm) < Version(right_norm))
     except Exception:  # noqa: BLE001
-        return None
+        # PEP 440 (``packaging.Version``) rejects npm-style pre-release tags
+        # like ``13.4.20-canary.13`` / ``5.0.0-rc.1`` / ``1.0.0-beta.4`` which
+        # are valid SemVer for npm/yarn/pnpm publishes. Without a fall-back
+        # the comparator returns None, the OSV/GHSA range matcher conserva-
+        # tively marks the package as affected, and downstream emits a false
+        # positive (e.g. ``next@16.2.4`` flagged by an advisory whose fix is
+        # ``< 13.4.20-canary.13``).
+        #
+        # Retry with the pre-release suffix stripped from BOTH sides so we
+        # get a defensible numeric major.minor.patch comparison. This is
+        # technically lossy (``13.4.20-canary.X`` is a pre-release of
+        # ``13.4.20`` per SemVer 2.0), but for OSV/GHSA introduced/fixed
+        # bounds the major.minor.patch view is the safe answer: operators
+        # on a strictly higher major.minor.patch shouldn't be flagged.
+        try:
+            from packaging.version import Version
+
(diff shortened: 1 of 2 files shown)


msaad00:
auto-retrigger: required CI did not attach for 'Lint and Type Check' on head ac315d2108a9c16561dacfe1c78dfeb5755cec79 (likely a GITHUB_TOKEN-authored 'Update branch' merge or cancelled required context). Reopening to re-run checks.

msaad00:

agent-bom (approved):

andres-linero (approved):

PR #2253, opened 2026-05-05
fix(ui): /graph readability — collapse metadata, dock legend, reclaim viewport

msaad00:
## The first-impression problem

Every time an operator opened \`/graph\`, they saw ~700px of stacked metadata before the React Flow canvas:

- 4-card \`SnapshotMetaCards\` row (Snapshot / Topology / Scope / Window) repeating the same values that the inline summary already showed
- 5-card Snapshot Diff block — useful for diffing scans, never useful on first paint
- 4-bullet "How to read this graph" prose
- Entity legend hundreds of pixels away from the canvas it documented

On a 1080p laptop the actual graph viewport was below the fold. The user feedback in the audit thread: *"too much going on"* / *"any person looks at this will not benefit, too crowded"*. Fair.

## Changes

\`ui/app/graph/graph-page-client.tsx\`

| Before | After |
|---|---|
| 4 \`SnapshotMetaCards\` row (Snapshot/Topology/Scope/Window) | **Removed** — inline summary already surfaces nodes / edges / captured-at / paging |
| 5-card Snapshot Diff block (~250px vertical) | One-line collapsible: \`+45 −11 nodes · +59 −11 edges · show\` |
| 4-bullet "How to read this graph" prose | One-line collapsible \`<details>\` |
| \`<GraphLegend>\` hundreds of px away in hero top-right | Docked onto canvas via \`<Panel position="top-right">\`, collapsible |

## Net effect

Same data, ~70% more vertical real estate for the canvas on first paint. Every collapsed block still expands to its full prior content with one click. Legend now lives next to the colors it documents. No behaviour change — pure layout / hierarchy.

## Verification

- [x] \`npm run typecheck\` clean
- [x] Captured live screenshot against /graph with 244-node self-scan: canvas + filter sidebar + first finding cards now in the SAME fold that previously held only metadata
- [ ] CI green

## Why this gates v0.86.0

The graph IS the moat. Shipping the tag with the moat surface looking like a wall of metadata cards on first paint torches every enterprise reviewer's first impression. This is a 100-line UI cleanup, no API changes, no behaviour changes — it's worth the 30 min in the release.

### DIFF
--- a/ui/app/graph/graph-page-client.tsx
+++ b/ui/app/graph/graph-page-client.tsx
@@ -761,17 +761,6 @@ export default function GraphPageClient() {
     graphData && graphData.pagination.limit > 0
       ? Math.max(1, Math.ceil(graphData.pagination.total / graphData.pagination.limit))
       : 1;
-  const relationshipScopeLabel =
-    filters.relationshipScope === "all"
-      ? "all relationships"
-      : `${filters.relationshipScope} relationships`;
-  const runtimeModeLabel =
-    filters.runtimeMode === "all"
-      ? "static + runtime"
-      : filters.runtimeMode === "static"
-        ? "static only"
-        : "runtime only";
-
   if (loadingSnapshots) {
     return (
       <div className="flex items-center justify-center h-[80vh] text-zinc-400">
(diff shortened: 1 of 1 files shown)


agent-bom (approved):
✅ Approved — UI cleanup verified, typecheck + lint clean, captured screenshot confirms snapshot-diff and how-to-read are now collapsed default and legend is docked on the canvas.

andres-linero (approved):

PR #2312, opened 2026-05-06
ci: pin uv setup tool version

msaad00:
Summary
- Pins the uv binary version used by the shared Python setup action.
- Keeps the setup action itself SHA-pinned while avoiding live latest-release lookup during CI startup.

Why
- The setup-python dependency bump surfaced a failure mode where GitHub API instability prevented uv from installing, which made later tests fail with `uv: command not found`.

Validation
- python scripts/check_workflow_timeouts.py
- git diff --check


agent-bom (approved):
Reviewed CI-only setup action pin.

andres-linero (approved):

msaad00:
Superseded by #2315, which carries the same uv setup pin as a verified signed commit.

PR #2540, opened 2026-05-12
fix check output contract gaps

msaad00:
Summary
- Add `agent-bom check --format sarif` so pre-install package gates can emit SARIF directly.
- Preserve vulnerability enrichment metadata in check SARIF: CVSS security severity, EPSS, KEV, CWE, aliases, advisory sources, and fixed version.
- Expose `packages` as a top-level JSON alias for `ai_bom_entities.packages` so consumers do not see an empty or missing package surface while `summary.total_packages` is nonzero.
- Keep `--agent-mode` constrained to the JSON envelope and reject SARIF there instead of mixing contracts.

Root Cause
The package check path had only console/JSON output while the rest of the scanner surfaces supported SARIF. Scan JSON also required consumers to know that package entities lived under `ai_bom_entities.packages`, which made the top-level package contract look empty or drifted from summary counts.

Validation
- `uv run pytest -q tests/test_cli_check.py tests/test_output_formats.py`
- `uv run ruff check src/agent_bom/cli/_check.py src/agent_bom/output/json_fmt.py tests/test_cli_check.py tests/test_output_formats.py`
- `git diff --check`
- commit hooks: ruff, ruff-format, mypy, bandit, env-var drift, duplicate artifact guard

### DIFF
--- a/src/agent_bom/cli/_check.py
+++ b/src/agent_bom/cli/_check.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import json
 import sys
 from contextlib import nullcontext
+from hashlib import sha256
 from pathlib import Path
 from typing import Any, Optional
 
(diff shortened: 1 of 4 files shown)


agent-bom (approved):
Approved: scoped check output contract hardening with regression coverage.

agent-bom (approved):
Approved after check retrigger commit; scope unchanged.

andres-linero (approved):

PR #2714, opened 2026-05-20
fix(discovery): accept project config files

msaad00:
Summary
- accept direct project MCP config file paths wherever project-scoped discovery is used
- keep Docker Compose and dynamic discovery scoped to the config file parent directory instead of treating the file as a directory
- add a regression for direct .cursor/mcp.json discovery

Verification
- uv run pytest tests/test_project_mode.py::test_discover_all_accepts_project_config_file_scope -q
- uv run ruff check src/agent_bom/discovery/__init__.py tests/test_project_mode.py
- uv run mypy src/agent_bom/discovery/__init__.py
- python scripts/check_release_consistency.py
- git diff --check

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -680,13 +680,13 @@ jobs:
       - name: Install build deps (Alpine)
         run: |
           for attempt in 1 2 3; do
-            if apk add --no-cache bash gcc g++ musl-dev libffi-dev openssl-dev git curl; then
+            if apk add --no-cache bash gcc g++ musl-dev libffi-dev openssl-dev git curl tar; then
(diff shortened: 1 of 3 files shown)


msaad00:
Root cause: the Alpine test suite passed, but the post-job uv cache save failed because BusyBox tar does not support --posix. This installs GNU tar in the Alpine job so setup-uv cache save can complete. Verification: uv run pytest tests/test_project_mode.py::test_discover_all_accepts_project_config_file_scope -q; python scripts/check_release_consistency.py; git diff --check.

andres-linero (approved):

agent-bom (approved):

PR #2797, opened 2026-05-26
fix(ci): retry pip-audit with pypi source

msaad00:
Summary
- retry pip-audit with the PyPI advisory source when the OSV source crashes before writing JSON
- apply the same fallback to current and base audit reports so the delta gate still compares artifacts

Verification
- python - <<'PY'
from pathlib import Path
import yaml
yaml.safe_load(Path('.github/workflows/pr-security-gate.yml').read_text())
print('workflow yaml parses')
PY
- git diff --check

Notes
- GitHub CI showed pip-audit crashing in its OSV adapter with KeyError: ranges before writing audit.json; this keeps the security gate fail-closed only when both advisory sources fail to produce a report.

### DIFF
--- a/.github/workflows/pr-security-gate.yml
+++ b/.github/workflows/pr-security-gate.yml
@@ -88,16 +88,77 @@ jobs:
 
       - name: Run pip-audit on current checkout
         run: |
-          set +e
-          uv run --with pip-audit pip-audit --ignore-vuln CVE-2026-3219 \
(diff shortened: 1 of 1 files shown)


agent-bom (approved):
CI fallback is scoped to pip-audit source resilience.

agent-bom (approved):

andres-linero (approved):

PR #2886, opened 2026-06-05
feat(gateway): act on cost-spike anomalies (detection → enforcement)

msaad00:
## What

Opt-in **anomaly-triggered enforcement** at the gateway, the cost counterpart to drift enforcement. When an agent's spend is a statistical outlier vs the tenant fleet (cost-spike anomaly), it can be blocked or flagged at relay time via `--anomaly-enforcement` (env `AGENT_BOM_GATEWAY_ANOMALY_ENFORCEMENT`):

- `enforce` — the anomalous agent is **blocked** (JSON-RPC `-32001`) with an `anomaly.blocked` governance event
- `warn` — the call relays but is audited as `gateway.anomaly_warned`
- `off` (default) — unchanged; anomalies stay advisory

## Why

Cost-anomaly detection already flagged runaway agents, but the signal was advisory-only — the agent kept relaying until it exhausted an absolute budget, if one was even set. Anomaly enforcement catches a spike *relative to the agent's peers* even while still under budget, closing the other half of the runtime/cost governance gap (the first half was drift enforcement).

## How

Backed by `cost_anomalous_agents()`, a per-tenant **TTL cache (30s)** over the existing modified-z detection, so the hot relay path never re-aggregates spend per call. The gate runs *after* the absolute-budget check (so it adds the relative-spike signal on top), is keyed on the resolved `source_agent`, and **fails open** — a cost-store error logs and allows rather than breaking the relay. Default-off keeps enabling it an explicit operator decision.

## Tests

`tests/test_gateway_anomaly_enforcement.py`: enforce-blocks-runaway, normal-agent-allowed, warn-audits-without-blocking, off-is-advisory, store-failure-fails-open. Gateway + finops suites green; ruff / ruff-format / mypy / bandit clean.

### DIFF
--- a/scripts/env_var_allowlist.txt
+++ b/scripts/env_var_allowlist.txt
@@ -85,6 +85,8 @@ AGENT_BOM_EXTERNAL_SECRETS_ENABLED
 AGENT_BOM_GATEWAY_ALLOW_INSECURE_NO_AUTH
 AGENT_BOM_GATEWAY_BEARER_TOKEN
 AGENT_BOM_GATEWAY_DETECT_VISUAL_LEAKS
+# opt-in anomaly-triggered gateway enforcement mode (off|warn|enforce)
+AGENT_BOM_GATEWAY_ANOMALY_ENFORCEMENT
 # opt-in drift-triggered gateway enforcement mode (off|warn|enforce)
 AGENT_BOM_GATEWAY_DRIFT_ENFORCEMENT
 # shared fallback policy path for standalone gateway and control-plane firewall decision route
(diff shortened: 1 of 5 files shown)


msaad00:
Updated and pushed: rebased onto current main, replaced the public anomaly block reason with a safe policy-source response, and resolved the stale GitHub Advanced Security review thread. All checks are green; merge is now blocked only by required code-owner review from @andres-linero.

agent-bom (approved):

andres-linero (approved):

PR #3005, opened 2026-06-22
ci: tolerate async external rebuilds in publish-registries

msaad00:
## Why
On the v0.89.2 release, `publish-registries` red-failed on two post-publish verification steps even though the catalog pushes (Smithery + ClawHub) and PyPI/Docker/GitHub Release all succeeded:

- **Smithery** — "Validate Smithery public endpoint" probes `/health` with `--forbid-auth-required`. The Smithery-hosted server rebuilds **asynchronously** and is **auth-gated** (`/mcp` → 401, no public `/health`) — which is the correct secure posture, not a failure. The probe can never pass for the deployed endpoint, so it killed the run.
- **Glama** — "Verify Glama listing freshness" allows ~3 minutes (6×30s), but Glama re-crawls listings on its own cadence (hours). It legitimately fails right after the rebuild trigger.

These false failures are exactly the "why did the publish fail" noise that obscures real problems.

## What
Make both post-publish external-rebuild verifications **best-effort** (`continue-on-error: true` + `::warning::` with a self-resolve note + the listing URL). The catalog push steps are unchanged, so a genuine publish failure still fails; only the async-rebuild liveness/freshness probes degrade to warnings.

Verified: YAML valid, actionlint clean.

## Follow-up (not in this PR)
A scheduled freshness re-check that opens an issue if a registry is *still* stale after a longer grace window would convert these warnings into real alerts — the "self-healing + notify" goal — without gating release day.

### DIFF
--- a/.github/workflows/publish-registries.yml
+++ b/.github/workflows/publish-registries.yml
@@ -67,30 +67,44 @@ jobs:
           fi
           echo "enabled=true" >> "$GITHUB_OUTPUT"
 
+      # Post-publish liveness probe of the Smithery-hosted endpoint. Best-effort
+      # confirmation, NOT a gate: Smithery rebuilds the server asynchronously
+      # (the new version may not be live for minutes) and the hosted endpoint is
+      # auth-gated with no public /health — both the correct secure posture, not
+      # a failure. A hard failure here previously red-failed the whole publish
+      # even though the catalog push (next step) succeeded. Warn instead so
+      # staleness stays visible without burning the run.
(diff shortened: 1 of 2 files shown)


msaad00:
auto-retrigger: required CI is stranded on head e89545be2477cf0b382b3959bc28ebbf7b694a70 (missing: Lint and Type Check). Reopening to re-run checks.

msaad00:

andres-linero (approved):

agent-bom (approved):

PR #3011, opened 2026-06-22
test: prevent xdist auth-env leak (intermittent 401!=200 flake) for good

msaad00:
## The recurring flake
Every PR has been intermittently blocked by `assert 401 == 200` on API route tests (`test_snowflake_backend_contract::test_supported_snowflake_routes_remain_wired`, `test_context_graph::TestAPIContextGraph::*`, etc.), requiring manual reruns. This kills it at the root.

## Root cause
The autouse reset in `conftest.py` calls `configure_api(api_key=None)` to return the API to its open state. But `configure_api` **re-derives** whether the API-key middleware is installed from `os.environ` (`auth_configured = api_key or env_keys or oidc_enabled or trusted_proxy_enabled or scim_enabled` — all read live from env). The conftest cleared **zero auth env vars**, so if a prior test left one set — e.g. `AGENT_BOM_TRUST_PROXY_AUTH` via `tests/auth_helpers.enable_trusted_proxy_env()` or a `setup_module` block whose teardown was skipped on a worker crash/error — the reset itself **re-installs `APIKeyMiddleware`**, and the next test's header-less GET returns 401 instead of 200.

## Fix (safe + preventive)
Snapshot the auth-env surface in the autouse fixture's **setup** (which runs *after* `setup_module`, so module-scoped auth env is captured and preserved across that module's tests) and **restore it on teardown** — reverting any auth var a *test body* set and failed to clean up, even on an assertion error. Leaks become impossible regardless of which test causes them.

## Proof
- Repro (a test that enables auth + skips cleanup, then an open test): **fails on `main`** (`'1' is None`), **passes with this fix**.
- Module-scoped auth suites still pass: 62 in proxy/oidc/scim + 120 across the previously-flaky modules.
- New permanent regression test `tests/test_auth_env_isolation.py`.

Test-only change. No production code touched.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -188,6 +188,26 @@ def _reset_durable_store_singletons() -> None:
         pass
 
 
+# Auth-influencing env vars. ``configure_api`` re-derives whether the API key
+# middleware is installed from these on every call (server.py auth_configured),
+# so any one of them left set makes an otherwise-open endpoint return 401. A
+# test that sets one directly (tests/auth_helpers.py, or a setup_module block)
+# and then errors before its own cleanup runs would leak it to the next test on
+# the same xdist worker — the root of the intermittent ``assert 401 == 200``
+# flake. We snapshot these before each test and restore them after; see below.
+_AUTH_ENV_VARS = (
+    "AGENT_BOM_API_KEY",
+    "AGENT_BOM_API_KEYS",
+    "AGENT_BOM_TRUST_PROXY_AUTH",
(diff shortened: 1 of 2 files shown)


msaad00:
auto-retrigger: required CI is stranded on head 946655f3a92f70e6274432726587ad663399a924 (missing: Lint and Type Check). Reopening to re-run checks.

msaad00:

andres-linero (approved):

agent-bom (approved):

PR #3095, opened 2026-06-24
feat(cloud): add Snowflake inventory gate symmetric with AWS/Azure/GCP

msaad00:
## Summary

Snowflake estate discovery was reachable only via the \`--snowflake\` CLI flag, unlike AWS/Azure/GCP which each expose an \`AGENT_BOM_<PROVIDER>_INVENTORY\` env flag plus an \`inventory_enabled()\` function. This adds the symmetric Snowflake gate so cloud-enablement is consistent across every provider.

## Changes

- **\`cloud/snowflake.py\`**: add \`INVENTORY_ENV_FLAG = "AGENT_BOM_SNOWFLAKE_INVENTORY"\` and \`inventory_enabled()\` (read-only, no side effects, same \`_TRUTHY\` pattern as the other providers), plus a shared \`enrich_report_with_snowflake_estate(report)\` helper that runs all nine estate discoveries (object graph, login anomalies, exfil, auth posture, services, pipeline, integrations, external data, governance, activity) and attaches the \`snowflake_*_data\` blocks. Each discovery is best-effort; a connector failure never breaks a scan.
- **\`scan_enrichment.py\`**: run the shared helper from \`enrich_report_with_estate_discovery\` when the gate is on. The Snowflake estate produces distinct \`snowflake_*_data\` blocks rather than \`collect_cloud_inventory\`'s single inventory-dict shape, so it attaches via the parallel estate helper. Default OFF and crash-safe.
- **\`cli/agents\`**: route the \`--snowflake\` path through the same shared helper, removing the duplicated per-discovery logic so the CLI flag and the env gate run identical code.
- **\`scripts/env_var_allowlist.txt\`**: declare \`AGENT_BOM_SNOWFLAKE_INVENTORY\`.
- **tests**: gate truthiness (default-off / on / variants) + enrichment no-op-when-off / runs-mocked-when-on / crash-safe.

## Verification

- \`uv run pytest tests/ -k "snowflake or scan_enrichment or inventory" -q\` → 588 passed, 1 skipped
- \`ruff format\` / \`ruff check\` / \`mypy\` clean; pre-commit env-var-drift passes.


msaad00:
Duplicate of #3094 (identical commit 2bd8615e). Closing to keep one PR for the Snowflake inventory gate.

PR #3108, opened 2026-06-24
fix(cloud): complete the gcp extra (compute/iam/storage/logging) + SDK guard

msaad00:
## Why
`gcp_inventory` imports `compute_v1` (instances + firewalls), `iam_admin_v1` (service accounts), `storage`, and `logging`, but the `gcp` extra declared none of them — only aiplatform/functions/container/run/resource-manager. On a clean `pip install 'agent-bom[gcp]'`, those imports raise ImportError and the entire compute/firewall/IAM/storage discovery class silently returns empty. Found while standing up the GCP POC.

## What
- Add `google-cloud-compute`, `google-cloud-iam`, `google-cloud-storage`, `google-cloud-logging` to the gcp extra.
- New `tests/test_gcp_sdk_coverage.py` — scrapes every `from google.cloud import X` in the GCP code and asserts each resolves under the installed extra (mirrors the boto3 / Azure SDK guards), so the extra can't drift incomplete again.

Verified: `compute_v1` + `iam_admin_v1` + `storage` all import after the fix.

### DIFF
--- a/src/agent_bom/cloud/aws_inventory.py
+++ b/src/agent_bom/cloud/aws_inventory.py
@@ -79,6 +79,10 @@ _AWS_IAM_PERMISSIONS: tuple[str, ...] = (
     "iam:ListUsers",
     "iam:ListAttachedRolePolicies",
     "iam:ListAttachedUserPolicies",
+    "iam:ListRolePolicies",
+    "iam:ListUserPolicies",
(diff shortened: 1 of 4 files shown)


msaad00:
Superseded by #3110 — this PR landed empty (commit aborted by a ruff E501 + the system-py3.9 atlas hook).

PR #3109, opened 2026-06-24
feat(cloud): auto-derive GCP project from ADC (parity with Azure subscription)

msaad00:
## Why
A GCP scan with a valid service-account key (or `gcloud` ADC) still returned `no_project` because the resolver only read `GOOGLE_CLOUD_PROJECT`. Operators expect the key/ADC to be enough — the SA key JSON already carries `project_id`. Same gap (and fix) as the Azure subscription auto-derive (#3105).

## What
- When no project is supplied, resolve it via `google.auth.default()`, which reads the project from the SA key's `project_id`, the gcloud config, or the metadata server — the same resolution every google client uses.
- Fail-safe: missing google-auth, ADC errors, or no project all degrade to a clear warning; never raises.

## Tests
New `tests/test_gcp_project_autoderive.py` (derive-from-ADC / no-project / ADC-error). GCP inventory suite green (14).

### DIFF
--- a/src/agent_bom/cloud/aws_inventory.py
+++ b/src/agent_bom/cloud/aws_inventory.py
@@ -79,6 +79,10 @@ _AWS_IAM_PERMISSIONS: tuple[str, ...] = (
     "iam:ListUsers",
     "iam:ListAttachedRolePolicies",
     "iam:ListAttachedUserPolicies",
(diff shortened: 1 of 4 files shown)


msaad00:
Superseded by #3110 — landed empty (commit aborted before the changes were captured).

PR #3145, opened 2026-06-25
Auto-refresh vulnerability cache and surface a freshness indicator

msaad00:
## Summary

Make the vulnerability data backing a scan demonstrably up to date: auto-refresh the local cache when it is stale or missing, and surface a structured freshness indicator that the CLI renders and the API/MCP tool return.

## What changed

- **Structured freshness snapshot** (`vuln_freshness.py`): a single object carrying the sources in play (OSV/GHSA/NVD/EPSS/KEV), `last_updated`, age, record count, and `stale`/`danger` flags. Age is computed from an injected reference time so the logic is deterministic and testable; reading it never opens the cache read-write and never raises.
- **Auto-refresh on staleness**: when a scan needs the vuln DB and the local cache is missing or older than `AGENT_BOM_VULN_DB_MAX_AGE_HOURS` (default 24h), the cache is refreshed before scanning. The refresh decision lives in one idempotent, threshold-respecting predicate, so a fresh cache is never re-synced. A refresh failure falls back to the live API or the existing cache and never blocks or fails the scan.
- **Offline / airgap preserved**: `--offline` and `AGENT_BOM_VULN_DB_OFFLINE` skip all network refresh and use only the existing cache; missing cache while offline is reported as a danger state rather than silently degrading.
- **Actionable indicator**: the bare "No local vulnerability DB found" warning is replaced with a line that states the source and age, e.g. `Vuln data: OSV+GHSA+NVD live (no local cache — run \`agent-bom db update\` for faster offline scans)`. A prominent warning fires when the cache crosses a clear-danger threshold (7 days) or the network is unreachable.
- **Cross-surface**: the same snapshot is attached to the report (`vuln_data_freshness`) and emitted in JSON output, so the API and MCP tool expose it; the CLI renders it. A new `db freshness` command prints the structured view (text or JSON) for CI and operators.
- Source labels are derived from the advisory-source catalog so they cannot drift; an injectable enrichment-cache age accessor adds a live-mode freshness signal.

## Tests

`tests/test_vuln_freshness.py` (new) and `tests/test_auto_update_db.py` (updated): missing-cache auto-refresh path, offline never refreshes, stale/danger thresholds, custom threshold via env, freshness object populated with sources + age, deterministic injected `now`, corrupt-DB graceful degradation. Touched files are ruff- and mypy-clean; the targeted `advisor/intel/vuln/freshness/db/enrich` suite is green (1147 passed).


msaad00:
auto-retrigger: required CI is stranded on head f6f7a9f27c62c25db7323fb813b4397fc93b1539 (missing: Lint and Type Check). Reopening to re-run checks.

msaad00:

andres-linero (approved):

agent-bom (approved):

PR #3159, opened 2026-06-26
feat(delivery): hardened outbound delivery foundation + OTLP traces emit

msaad00:
## Summary

Introduces one hardened outbound delivery foundation — `agent_bom.delivery.DeliveryClient` — that every exporter and webhook can ride, consolidating the previously scattered outbound logic (the shared `http_client` retry loop and the posture webhook outbox) behind a single, well-tested path.

## The foundation (`src/agent_bom/delivery.py`)

- **Retries + backoff + jitter + dead-letter** — exponential backoff with bounded attempts; on persistent failure the delivery is written to a durable dead-letter record and an actionable warning is returned. A destination failure never blocks or crashes the caller; the scan/relay continues.
- **Idempotency** — each delivery carries an idempotency key (a deterministic content hash by default). A retry, or a duplicate enqueue of the same payload to the same destination, is never double-sent; a key that already reached a terminal state returns a `deduplicated` result without hitting the wire.
- **HMAC request signing** (optional, per destination) and **per-destination auth** (bearer / token / custom header).
- **Circuit breaker** per destination — opens after a configurable number of consecutive failures, then half-opens a single probe after a cooldown; while open, deliveries short-circuit to dead-letter without a network call.
- **Delivery / audit log** — a queryable SQLite log records every attempt (destination, status, idempotency key, HTTP status, redacted payload preview). Secret values, signing secrets, and auth tokens are never persisted — only a fingerprint.
- **Redaction** — payloads are passed through the sensitive-payload sanitizer before any preview is logged.

Time, sleep, and jitter are injectable, so retry math and the audit log are fully deterministic under test.

## Routing + backward compatibility

`api/webhook_store.py` gains an additive `deliver_subscription_event` helper that routes a governance event through the foundation. The existing governance webhook outbox path and all default behavior are unchanged — the foundation uses its own opt-in `AGENT_BOM_DELIVERY_DB` (falling back to the shared DB, then a default path), so it does not hijack the existing posture outbox file. A regression test asserts default paths stay distinct and the existing subscription API surface is untouched.

## OTLP traces emit

`output/prometheus.py` already emits OTLP **metrics**. This adds `push_otlp_traces` — a `TracerProvider` + `OTLPSpanExporter` emit gated behind the existing OTLP endpoint config. It is opt-in and a graceful no-op when the endpoint is unset or the OpenTelemetry packages are absent (it does not fail a scan). A root `agent_bom.scan` span carries scan-level attributes with a child span per severity bucket. Combined with the existing trace ingestion, observability interop now works both ways.

## Tests

`tests/test_delivery_foundation.py` (24 cases + 2 optional-dep skips):

- `test_persistent_failure_retries_with_backoff_then_dead_letter` — retry → backoff (1, 2, 4s) → durable dead-letter
- `test_dead_letter_never_raises` — outbound failure degrades, never raises
- `test_idempotency_key_prevents_double_send`, `test_default_idempotency_key_is_deterministic_content_hash`, `test_deduplicated_after_dead_letter_is_not_delivered`
- `test_auth_failure_is_permanent_no_retry_and_warned`, `test_429_is_retried_and_warned`, `test_429_exhausted_warns_rate_limit`
- `test_circuit_breaker_opens_then_half_opens_probe`, `test_circuit_breaker_closed_on_success_resets_failures`
- `test_secret_values_not_in_delivery_log`, `test_signing_secret_never_logged`, `test_redacted_preview_strips_secrets`
- `test_hmac_signature_and_bearer_auth_headers`, `test_no_signature_header_without_secret`
- `test_webhook_store_delivers_through_foundation`
- `test_posture_outbox_path_default_unchanged_when_delivery_db_unset`, `test_existing_webhook_subscription_api_unchanged` — backward-compat
- `test_otlp_traces_noop_when_endpoint_unset`, `test_otlp_traces_noop_when_packages_missing`, `test_otlp_traces_emits_when_configured`, `test_otlp_traces_rejects_bad_endpoint`

`uv run ruff check` + `uv run mypy` clean on all changed files; full pre-commit green; `pytest -k "delivery or webhook or http or outbound or otlp or export"` → 281 passed, 3 skipped.

### DIFF
--- a/scripts/env_var_allowlist.txt
+++ b/scripts/env_var_allowlist.txt
@@ -104,6 +104,7 @@ AGENT_BOM_TOKEN_ROTATION_DAYS
 AGENT_BOM_CORS_ALL
 AGENT_BOM_DASHBOARD_CSP_HASH_MANIFEST
 AGENT_BOM_DB
+AGENT_BOM_DELIVERY_DB  # optional SQLite path for the outbound delivery foundation (audit log + idempotency ledger + circuit-breaker state); falls back to AGENT_BOM_DB then ~/.agent-bom/db/delivery.db; read in delivery.py
 AGENT_BOM_DB_PATH
 AGENT_BOM_DB_SOURCES
 AGENT_BOM_DEFAULT_ROLE
--- a/src/agent_bom/api/webhook_store.py
+++ b/src/agent_bom/api/webhook_store.py
@@ -294,6 +294,46 @@ def emit_governance_event(
         return 0
 
 
+def deliver_subscription_event(
+    subscription: WebhookSubscription,
+    *,
+    event_type: str,
+    payload: dict[str, Any],
+    idempotency_key: str = "",
+    client: Any = None,
+) -> Any:
+    """Deliver one event to a subscription through the hardened delivery
+    foundation (``agent_bom.delivery.DeliveryClient``).
+
+    Returns the ``DeliveryResult``. Best-effort and non-blocking by contract:
+    the underlying client never raises for a failed send — it dead-letters and
+    returns a warning. Callers that prefer the durable retry *worker* path
+    should keep using ``emit_governance_event`` (the posture outbox); this is
+    the inline, idempotent, circuit-broken alternative for direct delivery
+    (e.g. an operator "test" or a low-volume relay).
+    """
+    from agent_bom.delivery import Delivery, Destination, get_delivery_client
+
+    delivery_client = client or get_delivery_client()
+    destination = Destination(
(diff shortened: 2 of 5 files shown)


msaad00:
auto-retrigger: required CI is stranded on head 2ac479f60da4a14efa1c0266ef319700cfcae85f (missing: Lint and Type Check). Reopening to re-run checks.

msaad00:

agent-bom (approved):

andres-linero (approved):

PR #3162, opened 2026-06-26
feat(cli): capability doctor surface — every gated feature, state, and unlock path

msaad00:
## Why

agent-bom unlocks deeper coverage on stated conditions — an opt-in env var, a credential, a scoped cloud role, a prior scan, or data already present. For a security tool, a silently-skipped capability is the worst failure mode: a gap that looks like coverage. This adds a nothing-silent surface so every disabled, skipped, or degraded feature announces itself plus the exact path to turn it on.

## What

- **`src/agent_bom/capabilities.py`** — a declarative registry of gated capabilities. Each entry carries what it does, the unlock condition (env var / credential / scoped role / prior scan / data present), the env var names it reads (names only), and a probe that computes the live state — `on` / `off` / `degraded` / `unknown` — from the environment. Covers: per-provider cloud inventory (AWS/Azure/GCP/Snowflake), organization roll-up, agentless side-scan, audit-trail correlation, container-registry sweep, and the local vuln-DB cache + freshness.
- **`agent-bom capabilities`** — a grouped view (cloud / scan / runtime / data) showing each capability's state, why it is in that state, and a copy-pasteable unlock line for anything not enabled, ending with a one-line coverage summary.
- **`doctor`** — gains a capability section plus the one-line coverage summary, alongside the existing preflight checks.
- **`scan_status_line()`** — a reusable one-liner exposed for the scan path: "running with X enabled; Y available — set Z".

## Guarantees

- **Nothing-silent**: every gated feature lives in the registry; a test asserts each known gate appears with a state and an unlock path.
- **No secrets**: only the presence/absence of a variable name is ever reported — never a value. A test sets sentinel secret values and asserts they never appear in output.
- **Deterministic + graceful**: probes read only the environment and the local filesystem, never the network, and are wrapped so a failed probe degrades to `unknown` rather than crashing the command.

## Tests

22 tests in `tests/test_capabilities.py`: registry covers every known gate, unset flag → OFF + how-to, set flag → ENABLED, inventory opted-in-without-creds → DEGRADED, vuln-cache present/fresh logic, probe-never-raises, determinism, no-secret-values, and the CLI listing all capabilities with state + unlock. `ruff check`, `ruff format --check`, and `mypy` (new module added to the strict list) are clean.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -360,6 +360,7 @@ module = [
     "agent_bom.models",
     "agent_bom.backpressure",
     "agent_bom.proxy_sandbox",
+    "agent_bom.capabilities",
 ]
 disallow_untyped_defs = true
 disallow_incomplete_defs = true
--- a/site-docs/reference/cli.md
+++ b/site-docs/reference/cli.md
@@ -100,6 +100,7 @@ verbs are additive entry points that delegate to the underlying implementations.
 | `db freshness` | Show the structured vuln-data freshness indicator (sources, age, staleness) surfaced on every scan, API, and MCP tool |
 | `db framework-status` | Show bundled framework catalog freshness |
 | `doctor` | Check environment readiness for scanning |
+| `capabilities` | Show every gated capability, current state, and exact unlock path without printing secret values |
 | `gateway` | Multi-MCP gateway commands |
 | `interactive` | Start an interactive command shell for repeated CLI workflows |
 | `plugins` | Inspect extension plugin registry status without loading third-party plugin code |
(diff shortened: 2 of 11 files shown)


msaad00:
auto-retrigger: required CI is stranded on head 7cafdcfaf4839868c4cfb86d5bf94563a7b1ee4a (missing: Lint and Type Check). Reopening to re-run checks.

msaad00:

andres-linero (dismissed):

agent-bom (dismissed):

agent-bom (dismissed):

andres-linero (dismissed):

agent-bom (approved):

andres-linero (approved):

PR #3180, opened 2026-06-26
fix(ui): guard gateway live-feed KPIs against a partial payload

msaad00:
The gateway live-feed KPI cards verified that the `kpis` object existed but then called `.toLocaleString()` on individual fields. A partial KPI payload (object present, a field undefined) therefore crashed the whole gateway page through the route error boundary ("Something went wrong"), instead of degrading.

Fix: format every count through a small `fmtCount` helper that renders an em dash for a missing/non-numeric field rather than throwing. Each KPI card is now independently null-safe. The sibling `StatCard`/`FirewallStat` already render their value directly (no `.toLocaleString()`), so they were unaffected.

Adds a unit test for the formatter. typecheck + lint clean.


msaad00:
auto-retrigger: required CI is stranded on head dcc47e8dab4e0adf6194e92eb0db08ffdb3b1721 (missing: Lint and Type Check). Reopening to re-run checks.

msaad00:

agent-bom (approved):

andres-linero (approved):

PR #3201, opened 2026-06-27
feat(cloud): Vault key provider + MultiFernet rotation for connection encryption

msaad00:
Extends connection at-rest encryption (KMS work, #3197) with a **HashiCorp Vault provider** + **MultiFernet key rotation** — matching the Airflow/Superset convention for secure, rotatable secrets. Encrypt/decrypt API and the store are unchanged; only key resolution + Fernet construction.

- **Vault provider** (`AGENT_BOM_CONNECTIONS_KEY_PROVIDER=vault`): read-only `GET {AGENT_BOM_VAULT_ADDR}/v1/{mount}/data/{path}` with `X-Vault-Token`, key from KV v2 `data.data[field]` (`AGENT_BOM_CONNECTIONS_KEY_REF` = `mount/path#field`, field defaults to `key`). Uses the repo HTTP client (TLS verify on, no new dep). Cached + reset hook.
- **MultiFernet rotation** (all providers): any resolved key value may be a comma-separated list → `MultiFernet` that **encrypts with the primary (first) key, decrypts with any**. Single key stays plain Fernet (backward compatible). Documented zero-downtime procedure: add new key as primary → optional bulk re-encrypt → drop old.
- **Fail-closed / no-leak**: missing addr/token/ref, malformed ref, non-200, malformed body, missing field, transport error all raise `ConnectionSecretError`; errors carry only the failure mode — never the token, key, or Vault body (tests assert their absence).

New env vars (`AGENT_BOM_VAULT_ADDR`, `AGENT_BOM_VAULT_TOKEN`) + multi-key/vault notes in the allowlist. 54 tests (15 new: 9 Vault, 6 rotation); env-var drift OK; ruff/mypy(strict)/bandit clean.

### DIFF
--- a/scripts/env_var_allowlist.txt
+++ b/scripts/env_var_allowlist.txt
@@ -74,12 +74,16 @@ AGENT_BOM_CLICKHOUSE_URL
 AGENT_BOM_CLICKHOUSE_USER
 # Opt-in (default OFF) gate for the estate-wide read-only AWS asset inventory scanner.
 AGENT_BOM_CLOUD_INVENTORY
-# base64 Fernet key for at-rest encryption of cloud-connection secrets (ExternalId); secret, env/KMS-only, never in config.py or ENV_VARS.md. With provider=aws-kms it holds the base64 KMS-wrapped data key instead.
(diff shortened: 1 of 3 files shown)


msaad00:
auto-retrigger: required CI is stranded on head 831d6e6151566970d15dfada06ce8d72cca03342 (missing: Lint and Type Check). Reopening to re-run checks.

msaad00:

agent-bom (approved):

andres-linero (approved):

PR #3202, opened 2026-06-27
fix(ci): self-scan SARIF lands under its intended code-scanning category

msaad00:
Fixes the self-scan code-scanning uploads so each SARIF lands under its intended category instead of collapsing to `agent-bom` (which is why the Security tab showed a stale `agent-bom-scheduled` config and a blank-category analysis).

**Root cause:** agent-bom's SARIF formatter (`output/sarif.py`) hardcodes `runs[].automationDetails.id = "agent-bom/{scan_id}"`. `github/codeql-action/upload-sarif` only injects its `category:` input when `automationDetails` is absent, so the pre-set id won and GitHub derived the category as the substring before the final `/` → `agent-bom`. The empty `""` category came from the old legacy step setting a bare `agent-bom-scheduled` id (no slash → empty).

**Fix (workflow-only, all three uploads of formatter-generated SARIF):**
- `post-merge-self-scan.yml` — normalize (strip `automationDetails`) before the `agent-bom-self-dep` upload; removed the two dead legacy `agent-bom-scheduled` steps.
- `pr-security-gate.yml` — normalize before the real `agent-bom-self-scan` upload; removed the legacy `agent-bom-scheduled` placeholder. (The `agent-bom-self-dep`/`-self-sast` placeholders here use an empty SARIF with no `automationDetails`, so their categories already apply — left untouched.)
- `release.yml` — normalize immediately before the `agent-bom-release-gate` upload only; **no other job/needs/ordering changed** (verified; release call-graph linter passes).

`actionlint` clean on all three; pre-commit passes (the `release-workflow-lint` hook errored only on a missing-PyYAML env issue — verified it passes via direct run). After merge, the stale `agent-bom-scheduled` config can be deleted from the Security tab and won't reappear.

### DIFF
--- a/.github/workflows/post-merge-self-scan.yml
+++ b/.github/workflows/post-merge-self-scan.yml
@@ -126,34 +126,32 @@ jobs:
               print(f"... {len(findings) - 20} additional findings omitted from log")
           PY
 
-      - name: Upload dep scan SARIF
-        if: always() && hashFiles('sarif/self-dep.sarif') != ''
-        uses: github/codeql-action/upload-sarif@8aad20d150bbac5944a9f9d289da16a4b0d87c1e # v4
-        with:
-          sarif_file: sarif/self-dep.sarif
-          category: agent-bom-self-dep
-
-      - name: Prepare legacy scheduled SARIF category
+      - name: Align dep scan SARIF category
         if: always() && hashFiles('sarif/self-dep.sarif') != ''
         run: |
           python3 - <<'PY'
           import json
           from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


msaad00:
auto-retrigger: required CI is stranded on head f0f538bfbe62b2b41a377f0aaf7f73270ec7c856 (missing: Lint and Type Check). Reopening to re-run checks.

msaad00:

andres-linero (approved):

agent-bom (approved):

PR #3211, opened 2026-06-28
fix(scanner): align Debian image-CVE reporting + recover layered Python packages

msaad00:
## Summary

Two container-image scanning accuracy fixes in `image.py`, the OCI layer parser, and the vulnerability-matching/lookup layer.

### 1. False negatives — Python packages lost to overlay whiteout handling
The native OCI layer parser applied a **lower**-layer whiteout to files **re-created in a higher layer**. On Debian-based images the base layer whites out `site-packages` and a later layer reinstalls `pip`/`setuptools`/`wheel` at the same path, so all Python packages were silently dropped (0 pypi packages found; `CVE-2022-40897` for setuptools missed).

- Whiteouts are now applied as a **post-filter scoped to strictly-higher layers**, matching OCI overlay semantics: a path deleted by a higher layer is dropped; a path re-added by a higher layer survives a lower-layer whiteout.
- Added legacy `*.egg-info/PKG-INFO` / `*.egg-info/METADATA` extraction alongside modern `*.dist-info/METADATA` (de-duplicated) in both the native OCI parser and the `docker export` fallback, so older base images that ship setuptools/pip as egg-info are covered.

### 2. Over-reporting — distro CVEs not honoring the scanned release
On the same image agent-bom emitted **~845** CVE IDs vs the mainstream-scanner baseline of ~68. Root causes, all in the matching layer:

1. **Distro never detected.** `/etc/os-release` is a symlink to `/usr/lib/os-release`, which the parser refuses to follow, so distro came back empty and OS packages were matched against the fallback set of *every* Debian release (11/12/13/14) instead of the image's actual release. The parser now also reads `usr/lib/os-release`, pinning packages to e.g. Debian 10.
2. **Version comparator fed release-suffixed ecosystems.** The local DB stores distro ecosystems as `debian:10`/`alpine:v3.18`; the ecosystem-aware comparator only understands the base families, so every range collapsed to "unknown" and already-fixed CVEs were conservatively reported (e.g. `bash 5.0-4` flagged against a fix of `4.3-9.1`). DB ecosystems are normalized to `deb`/`apk`/`rpm` before comparison.
3. **Unfixed distro advisories surfaced.** OS-package advisories with no fix for the scanned release (no-dsa / won't-fix / end-of-life open) are now **suppressed by default**, aligning OS-package reporting with mainstream scanner conventions.

### Default suppression rule
For OS/distro ecosystems only (`deb`/`apk`/`rpm`): a distro advisory is suppressed by default when it has **no fixed version for the scanned release**. Distro security trackers assign a fixed version only once a patched package ships for that release; an entry left without a fix is the tracker's no-dsa / won't-fix / unimportant / end-of-life-open verdict. Application-ecosystem advisories (PyPI, npm, …) are never suppressed — an unfixed app CVE is still actionable. Opt back in with `AGENT_BOM_INCLUDE_UNFIXED=1` or `agent_bom.scanners.set_include_unfixed(True)`.

### Before / after (`python:3.9-slim-buster`)
| | Unique CVE IDs | Python pkgs | `CVE-2022-40897` (setuptools) |
|---|---|---|---|
| Before | ~845 | 0 | missed |
| After (default) | ~24 (buster-correct) | pip / setuptools / wheel | reported |

The post-fix count is limited by the bundled DB's Debian-10 coverage (buster is EOL); the previous ~845 were overwhelmingly already-fixed or wrong-release matches. The bulk reduction comes from correct release detection + comparison; the unfixed-suppression default additionally trims no-dsa/won't-fix noise (most visible on non-EOL releases).

### Tests
- egg-info extraction (OCI parser + `_packages_from_tar` fixture)
- whiteout re-add survives vs higher-layer deletion drops
- `usr/lib/os-release` distro detection
- release-suffixed ecosystem normalization + ordering; already-fixed not reported, genuinely-vulnerable reported
- unfixed-advisory suppression default + opt-in (setter and env var); fixed and app-ecosystem advisories kept

Read-only scanner behavior is unchanged.

### DIFF
--- a/docs/operations/ENV_VARS.md
+++ b/docs/operations/ENV_VARS.md
@@ -142,6 +142,11 @@ so they cannot regress silently, but they are not part of this reference.
 | `AGENT_BOM_MCP_MAX_TOOL_METRICS` | `int` | `128` | — |
 | `AGENT_BOM_MCP_TOOL_TIMEOUT_SECONDS` | `float` | `30.0` | — |
 
+## OS-package reporting
+| Env var | Type | Default | Description |
+|---|---|---|---|
+| `AGENT_BOM_INCLUDE_UNFIXED` | `bool` | `False` | When False (default), OS/distro advisories with no fix for the scanned release (no-dsa / won't-fix / end-of-life open) are suppressed so container reporting matches mainstream scanner conventions. Set AGENT_BOM_INCLUDE_UNFIXED=1 to surface  |
+
 ## PostgreSQL Control Plane Tuning
 | Env var | Type | Default | Description |
 |---|---|---|---|
--- a/src/agent_bom/config.py
+++ b/src/agent_bom/config.py
@@ -55,6 +55,16 @@ def _str(env_key: str, default: str) -> str:
 ENABLE_EXTENSION_ENTRYPOINTS = _bool("AGENT_BOM_ENABLE_EXTENSION_ENTRYPOINTS", False)
 
 
+# ── OS-package reporting ──────────────────────────────────────────────────────
+# When False (default), OS/distro advisories with no fix for the scanned release
+# (no-dsa / won't-fix / end-of-life open) are suppressed so container reporting
+# matches mainstream scanner conventions. Set AGENT_BOM_INCLUDE_UNFIXED=1 to
+# surface them. The scanner re-reads this env var at scan time so it can be
+# toggled per-invocation; see agent_bom.scanners.set_include_unfixed.
+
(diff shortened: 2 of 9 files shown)


msaad00:
auto-retrigger: required CI is stranded on head 80ebc71a6f0bc2a414176eda52611fea2df89453 (missing: Lint and Type Check). Reopening to re-run checks.

msaad00:

andres-linero (approved):

agent-bom (approved):

PR #3213, opened 2026-06-28
feat(db): ingest Debian Security Tracker for accurate per-release Debian coverage

msaad00:
## Why

agent-bom enriches OS packages from OSV, which **drops Debian releases once they reach end-of-life**. On `python:3.9-slim-buster` (Debian 10) OSV's bulk export has effectively no `debian:10` rows, so 17 real, fixable CVEs went undetected. A naive "nearest-release fallback" (querying `debian:11..14` for a buster image) would catch them but manufacture hundreds of false positives, because OSV cannot see buster's backported (`+deb10uN`) fixes. The dpkg comparator and per-release scoping were already correct — the gap was a missing data source.

## What

Ingest the **Debian Security Tracker** as an authoritative per-release data source.

- **New `debian` DB source** (in the default sync and `agent-bom db update --source debian`). It parses two feeds with an identical JSON shape through one parser:
  - the main tracker — supported releases (bullseye…forky), and
  - the Extended LTS tracker — the **end-of-life releases (buster/Debian 10)** the main feed has pruned, including their `+debNuX` backports.
- Per source package and release, status is mapped to a row under `debian:10..14`:
  - `resolved` → affected row bounded by the **per-release backported fix**,
  - `open` → no-fix row (no-dsa / won't-fix / EOL) — **suppressed by default**, restored with `AGENT_BOM_INCLUDE_UNFIXED=1`,
  - `undetermined` / `not-affected` / `fixed_version: "0"` → skipped (no false positives).
- Rows reuse OSV's `DEBIAN-CVE-*` id: `INSERT OR IGNORE` on `vulns` preserves OSV severity/CVSS; resolved `affected` rows override so the **authoritative backport wins** per release; OSV stays the fallback where the tracker has no entry.
- **Lookup fix:** the reported fixed version now comes from the per-release `affected.fixed` for distro ecosystems, not the cross-release `vulns.fixed_version` rollup. The rollup made a no-fix-for-this-release entry look fixed (with a wrong-release version) and leaked it past the default unfixed-advisory suppression.
- Image `dpkg` parsing now captures the `Source:` field so binary packages resolve to their source name (`libc6` → `glibc`) for the source-keyed tracker/OSV rows.
- ELTS is best-effort: a feed outage degrades only EOL coverage and never aborts the sync.

## Validation

Built the scanner from this branch, synced the `debian` source, and compared default output to the Debian-tracker baseline on three images. Every reported finding was re-checked with the dpkg comparator (`installed < fix`).

| Image | Release | baseline fixable | abom default (CVEs / findings) | missed baseline | installed≥fix FPs | `INCLUDE_UNFIXED` |
|---|---|---|---|---|---|---|
| `python:3.9-slim-buster` | debian 10 (EOL) | 26 | **91 / 189** | **0** | **0** | 246 |
| `python:3.11-slim-bookworm` | debian 12 | 6 | 7 / 7 | **0** | **0** | 82 |
| `python:3.12-slim` | debian 13 (trixie) | 4 | 5 / 5 | **0** | **0** | n/a |

- **Buster (the target):** before this change the default missed **17 of 26** fixable CVEs; now it misses **0**, with **zero** `installed≥fix` false positives.
- The buster default count (91 CVEs) is **above** the baseline's 26, but all extras are **genuine** `+deb10uN` backports the baseline's pre-ELTS buster data lacks (e.g. `libssl1.1` installed `…+deb10u5` vs fixed `…+deb10u8`, `libexpat1` `u6`→`u8`). Every finding was verified `installed < fix` — this is the "balloon to 400+ false positives" scenario *avoided*: 0 FPs, 0 unknown comparisons.
- **Bookworm / trixie (supported):** no regression — the full baseline fixable set is caught, totals do not explode, and `AGENT_BOM_INCLUDE_UNFIXED=1` works.

## Honest gaps / notes

- The literal `…/tracker/data/json` endpoint in the task no longer carries EOL buster (it was pruned when buster left standard support), which defeats the goal on its own. The implementation therefore also ingests the **Extended LTS** tracker feed (same JSON shape, run by the Debian LTS maintainers), which retains buster and its backports. This is the correct authoritative per-release source for EOL releases.
- The `Source:`-field capture in `image.py` is the deb-extraction path (not the python extraction the task scoped out) and is required for the source-keyed tracker rows to match binary packages.
- Sync downloads two large JSON feeds (~77 MB + ~100 MB); ingestion is batched/committed incrementally. No EOL-coverage-warning path was touched.

## Tests

`tests/test_db_sync_debian.py` covers release parsing, ingestion, OSV-merge precedence, best-effort ELTS, and EOL backport matching. Full DB/image/lookup/scanner suites pass; ruff/ruff-format/mypy/pre-commit clean.


msaad00:
auto-retrigger: required CI is stranded on head 3a979c800b5b6e28c7f0105d7e34ac72d15348a2 (missing: Lint and Type Check). Reopening to re-run checks.

msaad00:

agent-bom (approved):

andres-linero (approved):

PR #3244, opened 2026-06-29
fix(integrity): bind cosign verification to release identity

msaad00:
## Why
The release audit found instruction-file Sigstore verification accepted any signer because cosign was called with wildcard certificate identity and wildcard issuer. That makes a cryptographic verification look stronger than it is.

## What
- Bind cosign verification to the agent-bom release workflow identity by default.
- Require the GitHub Actions OIDC issuer by default.
- Keep explicit env overrides for controlled non-release/internal signing flows.
- Add regression tests that inspect the actual cosign command and reject wildcard trust.

## Tests
- uv run pytest tests/test_instruction_provenance.py -q
- uv run ruff check src/agent_bom/integrity.py tests/test_instruction_provenance.py
- uv run mypy src/agent_bom/integrity.py


msaad00:
Retriggering pull_request checks; reopening immediately.

agent-bom (approved):

andres-linero (approved):

PR #3256, opened 2026-06-29
fix(ui): restore client JS budget headroom after #3254 merge

msaad00:
PR #3254 (`fix(ui): wire cloud connection scan schedules`) merged ~425 B of additional client JS without bumping the checked-in UI bundle budget. As a result `totalClientJsBytes` (3,065,000) is now exceeded on main (actual 3,065,425), which fails the **UI bundle budget** step of UI Validate for every PR built off main.

This raises `totalClientJsBytes` to 3,075,000 with a documenting comment, matching the existing convention. Verified locally: total client JS 2993.6 KiB / budget 3002.9 KiB — PASS.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


msaad00:
Superseded by #3257, which restores bundle headroom by trimming the connection scheduler client code instead of raising the checked budget.

PR #3316, opened 2026-06-30
fix(release): harden post-0.90 audit findings

msaad00:
## Summary
- harden SAML assertion replay handling with atomic nonce consumption and bind exception audit actors to authenticated request identity
- normalize GHSA API ecosystems to scanner DB keys, preserve NVD incremental cursor on capped syncs, and make CPE range/vendor matching consistent in scalar and batch scan paths
- refresh stale 0.90.0 release references in docs/examples and keep CPE wording honest as opt-in review-grade matching

## Validation
- `uv run pytest -q tests/test_db_sync_ghsa.py tests/test_cpe_accuracy.py tests/test_db_sync_nvd_incremental.py tests/test_shared_auth_state.py tests/test_api_saml.py tests/test_api_enterprise_tenant.py -m 'not network'`
- `uv run ruff check src/agent_bom/api/shared_auth_state.py src/agent_bom/api/routes/enterprise.py src/agent_bom/cpe_match.py src/agent_bom/db/sync.py src/agent_bom/scanners/__init__.py tests/test_shared_auth_state.py tests/test_api_enterprise_tenant.py tests/test_cpe_accuracy.py tests/test_db_sync_ghsa.py tests/test_db_sync_nvd_incremental.py tests/test_online_scan_validation.py`
- `uv run mypy src/agent_bom`
- `uv run python scripts/check_release_consistency.py`
- `uv run python scripts/generate_env_var_reference.py --check`
- PyPI smoke: installed `agent-bom==0.90.0`, verified `agent-bom --version`, `agent-bom upgrade` from `0.89.2` to `0.90.0`, offline package scan, package check, IaC scan, and demo agent scan

## Notes
Local `uv` commands still warn about an old `.venv` `agent_bom-0.89.2.dist-info` missing `RECORD`; checks pass and PyPI smoke used clean venvs under `/tmp`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ credential environment names in reach, and the agents that can call it.
 
 | Domain | Coverage |
 |---|---|
-| Supply chain | Package and OS risk across npm, PyPI, Maven, Go, Cargo, NuGet, Composer, RubyGems, conda, Swift, Hex, Pub, apk, deb, and rpm, with OSV/GHSA/NVD enrichment, NVD CPE candidate matching for non-ecosystem software, transitive resolution, and dependency-confusion detection |
+| Supply chain | Package and OS risk across npm, PyPI, Maven, Go, Cargo, NuGet, Composer, RubyGems, conda, Swift, Hex, Pub, apk, deb, and rpm, with OSV/GHSA/NVD enrichment, opt-in review-grade NVD CPE candidate matching for non-ecosystem software, transitive resolution, and dependency-confusion detection |
 | Agents + MCP | MCP clients, servers, tools, transports, and trust posture with live introspection across 29 first-class client types |
 | AI models + datasets | Malicious-model detection via safe pickle-opcode disassembly (no deserialization), model/dataset cards, and target-scoped PII/PHI dataset-file scanning |
 | Cloud estate | Read-only, gated asset inventory and CIS posture across AWS, Azure, GCP, and Snowflake, plus AI/GPU provider posture |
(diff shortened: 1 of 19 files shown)


agent-bom (approved):

andres-linero (approved):

msaad00:
⚠️ **Blocking — verified HIGH regression found in post-merge audit (hold merge).**

The NVD incremental **capped-sync handling stalls and never makes forward progress** (empirically reproduced: 4 sequential capped runs, DB stays at 2 distinct CVEs, tail never ingested):

- On truncation the cursor is pinned to `start_ts` (`last_modified_end = start_ts`).
- But `_upsert_nvd_enrichment_row` returns `True` for **every** CVE with cvss/cwe (INSERT OR IGNORE + unconditional UPDATE), so `enriched` counts already-synced rows.
- `start_index` resets to 0 each run → every retry re-fetches page 0, re-saturates `enriched` at `max_results` on the **same head CVEs**, re-sets `sync_truncated`, cursor never advances.
- Any window with `totalResults > max_results` (default 120-day cold start routinely has thousands) re-syncs the head forever; the lookback clamp later slides `start_dt` forward and **silently skips** the unsynced tail.

This contradicts the change's own docstring ('retries it instead of permanently skipping') and is a regression vs main (which advanced the cursor). The capped test only asserts single-run truncation flags, so it passes while the stall persists.

Pushing a fix on this branch: advance the cursor to the max `lastModified` actually ingested (forward progress, overlap-not-skip) + a multi-run progress test.

PR #3320, opened 2026-06-30
fix(sca): make NVD capped sync ingest its unsynced tail across runs

msaad00:
## Summary
Fixes a **verified HIGH regression** introduced by #3316 and caught in the post-merge audit. The NVD incremental **capped-sync stalled and never made forward progress**, silently dropping CVEs — directly undermining NVD/CPE coverage (core to the product).

### Root cause
- On truncation the cursor is pinned to the window start.
- `_upsert_nvd_enrichment_row` returned `True` for **every** CVE with cvss/cwe, and `start_index` resets to 0 each run.
- So each run re-fetched page 0, re-saturated the cap on the **same head CVEs**, re-set `sync_truncated`, and never advanced. Empirically reproduced: 4 runs, DB stuck at 2 CVEs, tail never ingested; the lookback clamp later slid forward and skipped the unsynced tail.

### Fix
- `_upsert_nvd_enrichment_row` returns `True` only when it **inserts a new row**, so `enriched` counts genuinely-new work. A re-run pages past the synced head (no longer counted) and reaches the tail, draining the window across runs, then advances the cursor cleanly.
- Skip the trailing inter-page sleep once the cap is hit.
- Warn on truncation so an undersized `max_results` / oversized window is visible.

### Tests
New `test_sync_nvd_incremental_capped_runs_make_forward_progress` proves the tail is ingested across runs and the cursor advances on drain (the prior tests only checked single-run truncation flags, so the stall passed CI). Full `test_db_sync_nvd_incremental.py` green (4 passed); ruff clean.

Closes the audit's #3316 high finding.


msaad00:
Retriggering CI: branch is already current with main, so closing/reopening to fire fresh pull_request checks.

andres-linero (approved):

agent-bom (approved):

PR #3321, opened 2026-06-30
fix(graph): keep cross-page attack paths in filtered /graph

msaad00:
## Bug
The relationship/static/dynamic-filtered branch of `GET /graph` (`_filtered_graph_response`, `src/agent_bom/api/routes/graph.py`) silently dropped **every** derived attack path on graphs larger than the node page limit (default 500). Operators saw no kill-chains in filtered responses and wrongly concluded none existed.

## Root cause
The branch kept a derived path only when **all** of its hops were on the current node page:

```python
if p.hops and all(hop in paged_ids for hop in p.hops)
```

Derived paths span up to four nodes (agent -> server -> package -> vuln). On graphs above the page limit those hops land on different pages, so the `all(...)` check never held and the path disappeared. The store-backed branch already does this correctly: it selects paths whose **source** is on the page and back-fills off-page hop nodes via `nodes_by_ids`.

## Fix
Mirror the store branch in the filtered branch:
- Keep a path when its source hop is on the page (`p.hops and p.hops[0] in paged_ids`).
- Back-fill the off-page hop nodes (`{hop for kept path for hop in p.hops} - paged_ids`) into `nodes_by_id` from the in-memory graph so `_serialize_attack_path` resolves labels instead of emitting raw ids / `unknown` roles.

## Test
`tests/test_graph_filtered_attack_paths_paging.py` builds a graph larger than the page limit with the agent source on page 0 and the server/package/vuln hops pushed onto a later page. It asserts the filtered response still returns the attack path and that the off-page hop labels (`server:fs`, `vuln:CVE-1`) are resolved (correct labels, non-`unknown` roles). The test fails before the fix (path dropped) and passes after.

## Verification
- `python -m pytest tests/test_graph_filtered_attack_paths_paging.py tests/test_graph_attack_path_fusion.py tests/test_graph_attack_path_e2e.py -q` — 6 passed
- Confirmed the new test FAILS on pre-fix code (stashed fix), PASSES after
- `ruff check` on both changed files — clean


msaad00:
Retriggering CI: branch is already current with main, so closing/reopening to fire fresh pull_request checks.

agent-bom (dismissed):

andres-linero (dismissed):

andres-linero (approved):

agent-bom (approved):

PR #3325, opened 2026-06-30
fix(inventory): keep distinct MCP servers distinct across identity, enrichment, and Cortex audit

msaad00:
## The bugs

Three related inventory-correctness findings caused distinct assets to be merged, mis-enriched, or dropped during discovery.

**1. Identity collision (`discovery/identity.py`)** — `server_identity_key()` fell back to `cmd:{basename(command)}:{args}` for stdio servers. When a server had no args and no `url`/`registry_id`, every such server collapsed to `cmd:{command}:` — so two genuinely distinct command-less servers deduped into a single graph node.

**2. Enrichment maps to the wrong server (`mcp_introspect.py`)** — `enrich_servers()` built `result_map` keyed by `server.name` and matched each server by `server.name`. MCP server names are **not** unique across agents, so when two agents exposed servers with the same name, runtime tools/resources/prompts from one were merged onto the other.

**3. Cortex audit findings dropped (`discovery/__init__.py`)** — Cortex Code auxiliary security findings (`permissions.json` / `hooks.json` / `settings.json`) were attached only to an already-discovered `CORTEX_CODE` agent. If the `mcp.json` (and thus the agent) had not been seen yet — or did not exist — the findings were silently dropped. Behaviour depended on file-iteration order.

## Root cause

All three relied on the human-readable server/agent name (or an insufficiently distinctive command key) as an identity, and the Cortex path mutated shared state mid-loop instead of reconciling after discovery completed.

## The fix

- **identity:** a bare command with no args now folds the server name into the key (`cmd:{command}::{name}`), so distinct command-less servers stay distinct. Command+args dedup (process vs config) is unchanged.
- **enrichment:** `ServerIntrospection` now carries a stable `server_identity` (`server_identity_key`), populated wherever results are produced. `enrich_servers()` keys and matches by that identity, falling back to name only for results that predate identity tracking (keeps existing fixtures working).
- **Cortex:** pending Cortex metadata is buffered and flushed after the path loop — it attaches to the discovered `CORTEX_CODE` agent if present, otherwise surfaces on a standalone `CORTEX_CODE` agent, regardless of ordering.

## Tests

Added regression tests that fail before / pass after:
- `test_commandless_servers_with_same_command_stay_distinct` — two distinct command-less servers keep distinct identities and are not deduped.
- `test_enrich_servers_matches_by_identity_when_names_collide` — two agents sharing a server name each receive their own runtime data.
- `test_cortex_metadata_creates_standalone_agent_without_mcp` and `test_cortex_metadata_attaches_regardless_of_ordering` — Cortex findings surface with and without an `mcp.json`, independent of file order.

Ran: `pytest tests/test_discovery_identity.py tests/test_mcp_introspect.py tests/test_discovery_new_clients.py tests/test_introspect_cmd.py tests/test_core.py` (all green) and `ruff check` on all changed files (passed).


msaad00:
Retriggering CI after pushing mypy fix and branch refresh; closing/reopening to fire fresh pull_request checks.

agent-bom (approved):

andres-linero (approved):

PR #3356, opened 2026-07-01
fix(cli): make loopback first-run auth coherent (serve banner + quickstart)

msaad00:
## Problem

First-run loopback auth was incoherent:

- **P0** — `agent-bom serve`/`api` on loopback printed a `local unauthenticated mode (loopback only)` banner, but `/health` reported `unauthenticated_allowed: false` and `/v1/*` returned **401**. Root cause: the CLI passes `--allow-insecure-no-auth` (default `False`) as an *explicit* bool to `configure_api`, so the `allow_unauthenticated is None` branch never ran and `AGENT_BOM_ALLOW_UNAUTHENTICATED_API=1` was **silently ignored** on (re)configure (`api/server.py:721`).
- **P1** — `quickstart --run` (and the printed plan) told users to run `agent-bom serve --host 127.0.0.1 --port 8422` with no auth flag, so the cockpit it points at could not load data (401).

## Fix

- `configure_api` now ORs `AGENT_BOM_ALLOW_UNAUTHENTICATED_API` into the resolved posture on **every** (re)configure, so the env opt-out is honored on the CLI path. Non-loopback binds remain fail-closed — the CLI auth gate (`_enforce_auth_defaults`) refuses them before `configure_api` runs and does not consult the env var (verified: `serve --host 0.0.0.0` with the env var set still raises).
- The startup **Auth** banner is now derived from the same resolved inputs the server applies (`_api_auth_summary` + `_resolve_allow_unauthenticated`), so it can never claim unauthenticated access while the API fails closed. When **neither** flag nor env is set, it prints an accurate fail-closed message instructing `--allow-insecure-no-auth` / `--api-key` instead of the old misleading line.
- `quickstart` handoff (both `--run` and the printed plan) appends `--allow-insecure-no-auth` so the local demo cockpit loads, with a commented `--api-key <key>` alternative for shared hosts.

## Tests (fail-before / pass-after)

- `test_env_opt_out_honored_when_flag_passed_explicitly` — env opt-out honored when the flag is passed as explicit `False` (the exact serve/api call).
- `test_resolve_allow_unauthenticated_honors_env`, `test_api_auth_summary_*` — banner matches the real posture and is accurate when failing closed / prefers a configured API key.
- `test_{serve,api}_cmd_loopback_no_auth_reports_fail_closed` + `_env_opt_out_reports_unauthenticated` — banner reflects `/health`.
- `test_quickstart_run_*` — handoff includes a flag that lets `/v1/overview` load.

Gates: ruff, mypy (src/agent_bom), env-var/openapi/v1-schema/data-model/release-consistency `--check`, comment-hygiene, and the touched pytest suites all green. Did not touch `cpe_match.py` or `scanners/__init__.py`.


msaad00:
Retriggering checks: GitHub returned an empty statusCheckRollup for this mergeable PR, so closing/reopening to emit a fresh pull_request event without mutating the branch.

agent-bom (approved):

andres-linero (approved):

Issue #3350, opened 2026-06-30
epic (lane A): continuous / event-driven cloud posture


msaad00:
Move cloud posture from cadence polling (`scan_interval_minutes`) to near-real-time event ingestion so posture updates on resource change. Target ~60% of the realtime lane.
- [ ] PR A1: event-ingestion framework + AWS — consume CloudTrail/EventBridge change events (self-hosted: EventBridge→SQS, agent-bom long-polls the queue), map event→changed resource→inventory/graph delta→re-evaluate the affected CIS/posture rules; freshness indicator vs polling.
- [ ] PR A2: Azure (Activity Log / Event Grid) + GCP (Cloud Asset Inventory feed / Pub-Sub) event ingestion on the same framework.
- [ ] PR A3: orchestration — backpressure, dedup, idempotency, ordering, partial-failure handling; UI 'live/last-event' indicator; honest copy (event-driven where wired, polling fallback otherwise).

---
**Acceptance (every PR):** secure/fail-closed · bounded, no forever-loops · error-handled (graceful on provider/API failure + rate limits) · tested (fail-before/pass-after) · gate-clean (drift/openapi/data-model/bundle) · NO AI footer · NO competitor names. Consolidate related changes into substantial PRs (no micro-PRs).

PR #3358, opened 2026-07-01
feat(cloud): event-driven AWS posture — CloudTrail/EventBridge ingestion + per-resource rule re-eval

msaad00:
## What

Adds a **reactive, event-driven posture lane** for AWS alongside the existing polling scheduler. When a resource changes in a customer account, AWS emits a CloudTrail management event that EventBridge can route to SQS; the control plane drains that queue and re-evaluates **only the CIS rules for the changed resource's type** — so a bucket made public is re-checked in seconds instead of waiting for the next scheduled full scan. **Polling remains the fallback; this is additive.**

## How

- **`src/agent_bom/cloud/event_ingest.py`** (new)
  - `CloudChangeEvent` dataclass (provider, account, region, resource_type, resource_id, action, arn, raw) + `parse_cloudtrail_event()` that normalizes both the EventBridge envelope and a bare CloudTrail record.
  - `dispatch_change_event()` — resolves the changed resource via a **scoped** read-only inventory fetch (only the affected class), re-runs **only** the CIS checks for that resource type (`run_benchmark(checks=[...])`), and persists the resulting posture/delta through the **same scan/graph path** a full scan uses.
  - `consume_aws_events()` — a **bounded** SQS long-poll (operator wires EventBridge→SQS; opt-in via `AGENT_BOM_AWS_EVENT_QUEUE_URL`, default off). Processes at most a bounded batch then **returns** (no forever loop); visibility-timeout + delete-on-success; malformed/throttle → log + continue, never crash.
- **Fail-closed / secure:** read-only brokered role against the customer; drop any event whose account does not match the connection's own account (confused-deputy guard); bounded batches; malformed/foreign messages logged + dropped (poison), transient failures left on the queue for redelivery.
- **`last_event_at`** on the connection (event-driven-vs-polling freshness signal), threaded through the in-memory / SQLite / Postgres stores + schema.
- **Env vars:** 4 bounded-consume tunables declared in `config.py`; opt-in queue URL added to the allowlist; `docs/operations/ENV_VARS.md` regenerated.

## Tests

`tests/test_cloud_event_ingest.py` (15 tests): synthetic CloudTrail S3 `PutBucketPolicy` (bucket made public) re-evaluates the S3 CIS checks and produces a finding (CIS 2.1.1); malformed-event guard; foreign-account drop (dispatch + consumer); empty-queue no-op; bounded-batch stop; unknown-resource-type skip.

## Gates

- `generate_env_var_reference.py --check`, `export_openapi.py --check`, `generate_v1_schemas.py --check`, `regenerate_data_model_atlas.py --check`, `check_release_consistency.py` — all green (env-var reference regenerated + committed).
- `ruff check` clean; `mypy src/agent_bom` clean (602 files).
- `pytest` green for event ingest + connection store + scheduler (107 passed) and schema/parity/data-model sweep (63 passed).

## Deferred

- Always-on background driver that ticks `consume_aws_events()` per connection is intentionally **not** wired here (kept bounded / no-forever-loop per the trust posture). An operator or a scheduler tick invokes the consumer; wiring a periodic driver is a follow-up.

Part of #3350

### DIFF
--- a/docs/operations/ENV_VARS.md
+++ b/docs/operations/ENV_VARS.md
@@ -37,6 +37,10 @@ so they cannot regress silently, but they are not part of this reference.
 | `AGENT_BOM_API_SCAN_LEASE_SECONDS` | `int` | `600` | Distributed scan dispatch (multi-replica work-stealing). When enabled, scan jobs are enqueued to a shared Postgres dispatch queue and any control-plane replica claims them via FOR UPDATE SKIP LOCKED, so scan throughput scales with replicas  |
 | `AGENT_BOM_API_SCAN_WORKERS` | `int` | `min(4, os.cpu_count() or 2)` | — |
 | `AGENT_BOM_API_SCAN_WORKER_RECYCLE_JOBS` | `int` | `10` | — |
+| `AGENT_BOM_AWS_EVENT_MAX_BATCHES` | `int` | `10` | — |
+| `AGENT_BOM_AWS_EVENT_MAX_MESSAGES` | `int` | `10` | Event-driven AWS posture ingestion (continuous posture). When an operator wires EventBridge→SQS (opt-in via AGENT_BOM_AWS_EVENT_QUEUE_URL, read live, default off), the bounded SQS consumer drains change events and re-evaluates only the affe |
+| `AGENT_BOM_AWS_EVENT_VISIBILITY_TIMEOUT` | `int` | `120` | — |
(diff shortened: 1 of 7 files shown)


agent-bom (dismissed):

andres-linero (dismissed):

msaad00:
Retriggering pull_request checks for frozen/empty status rollup on current head.

agent-bom (approved):

andres-linero (approved):

PR #3364, opened 2026-07-01
feat(cli): zero-config loopback serve — auto-generate + print a dev API key

msaad00:
## What this does

#3356 made loopback `agent-bom serve` **fail closed** when no auth is configured — honest, but it forced every new user to pass `--allow-insecure-no-auth` (or wire a key) before the dashboard stopped returning 401. This removes that last friction **without weakening security**.

When `serve` binds to a **loopback** host (`127.0.0.1` / `localhost` / `::1`) **and** no auth path is configured (no `--api-key`, no `--allow-insecure-no-auth`, no `AGENT_BOM_API_KEY` / `AGENT_BOM_API_KEYS` / OIDC / SCIM / trusted-proxy, no `AGENT_BOM_ALLOW_UNAUTHENTICATED_API`), `serve` now:

1. Mints a single **ephemeral** `abk_…` dev API key (per-process, `secrets.token_urlsafe`).
2. Seeds it into the RBAC key store via `configure_api(api_key=…)` so the API accepts it as an admin key (`Authorization: Bearer …` / `X-API-Key`).
3. **Prints it prominently** in the startup banner: `Dev API key (loopback only): abk_… — the local dashboard uses this automatically`.
4. Auto-authenticates the same-origin dashboard: the HTML routes (`/`, SPA fallback) issue a matching **signed browser-session cookie** (admin, `key_id=None`, httponly + CSRF cookie). The UI loads with **zero flags** and **zero manual key entry**, keeping the existing header-free, cookie-based UI auth contract unchanged (no frontend changes needed).

## Security guarantees (fail-closed preserved)

- **Non-loopback binds NEVER auto-generate a key.** `_enforce_auth_defaults` still refuses unauthenticated non-loopback binds *before* key generation, and `_should_auto_generate_dev_key` independently re-checks the host.
- The dev key is **ephemeral** — per-process, never written to disk.
- Any explicit auth (key / flag / env / OIDC / SCIM / proxy) **suppresses** the dev key; explicit `--api-key` always wins.
- **`AGENT_BOM_NO_AUTO_DEV_KEY=1`** restores the prior fail-closed loopback behavior.
- The banner stays **honest**: it labels the key a loopback-only dev key, never claims unauthenticated access.
- The auto session cookie carries no `key_id` (so it doesn't depend on key-store lookups) and is reused across page loads instead of churning fresh nonces.

## Tests

[...]

PR #3369, opened 2026-07-01
fix(ci): stop order-dependent test failures — reset all runtime-written env vars

msaad00:
## Problem
Required Python test jobs fail **intermittently by random seed** (pytest-randomly shuffles collection order each run). Root cause: server/CLI runtime code sets **9** `AGENT_BOM_*` env vars as a side effect during test bodies (remote MCP bind, CORS, analytics/ClickHouse, profile), but the autouse reset fixture only snapshot/restored **auth + storage** vars. A test that triggers one of these paths and skips its own cleanup (e.g. on assertion error) leaks the value into later tests on the same worker.

Concretely: `AGENT_BOM_MCP_REMOTE_BIND=1` leaks into `tests/test_repo_scan.py`, whose `_remote_bind_active()` then demands `AGENT_BOM_REPO_SCAN_ALLOWED_HOSTS` and raises. Reproducible on **pristine main** with `pytest tests/test_cli_server.py tests/test_repo_scan.py` — main only stays green by lucky ordering. This is what's been reddening PRs (#3364 and others) unpredictably.

## Fix
Snapshot + revert the **full class** of runtime-written env vars in the autouse `reset_global_test_state` fixture, so outcomes are order-independent regardless of seed.

## Verification
- Previously-polluting pair now passes in **both** orders (42 passed).
- Broad ordering sweep (cli/repo_scan/core/api-hardening): **407 passed**.
- Full suite green.

Unblocks the required Python checks on all open PRs.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -212,6 +212,23 @@ _STORAGE_ENV_VARS = (
     "AGENT_BOM_POSTGRES_DSN",
 )
 
+# Env vars that server/CLI runtime code sets as a side effect during a test body
+# (e.g. binding the MCP server remotely, enabling permissive CORS). They are NOT
+# auth/storage config, but if a test that triggers this code path skips its own
+# cleanup the value leaks into later tests on the same worker and makes outcomes
+# depend on collection order — e.g. AGENT_BOM_MCP_REMOTE_BIND leaking into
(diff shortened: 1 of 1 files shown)


msaad00:
Retriggering pull_request checks for frozen/empty status rollup on current head.

agent-bom (approved):

andres-linero (approved):

PR #3397, opened 2026-07-01
fix(docs): differentiate workflow and architecture README diagrams

msaad00:
## Summary
- Redesign the control-plane architecture SVG as a vertical layered stack (sources → engine → evidence/platform → consumers) with a cool indigo/teal palette and LAYERED badge.
- Keep how-it-works as the horizontal five-lane workflow pipeline, framed with a warm amber accent and WORKFLOW badge so the two README visuals are immediately distinguishable.
- Regenerate light/dark SVG outputs from the updated generator.

## Test plan
- [x] `uv run python scripts/generate_doc_architecture_svgs.py`
- [x] `uv run pytest tests/test_doc_architecture_svgs.py -q`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,9 +165,9 @@ posture re-evaluation, not a full runtime CDR product.
 
 For DSPM, Snowflake has the deepest current support because agent-bom can read
 warehouse metadata, grants, tags, lineage, and governance activity visible to
-the configured role. AWS S3 also has an opt-in bounded content-sampling lane for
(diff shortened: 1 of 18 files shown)


msaad00:
Closing as superseded by current main. The diagram/script changes from this branch are already present on main, and the current PR head has an empty diff after merging main. The only remaining original README delta would regress the newer S3 + GCS DSPM wording, so there is nothing safe or useful left to merge/retrigger here.

Issue #3499, opened 2026-07-04
epic(scope): multi-language reachability and data-lake interop (audit T)


msaad00:
## Summary
Scope boundaries from audit: reachability analysis is Python-only today; data-lake interop (Iceberg export path) is light.

## Direction
- JS/TS/Go call-graph reachability (roadmap)
- Expand data-lake export/interop without making OCSF the internal model

## Related
- #2918 formatter/model convergence
- `docs/OCSF_BOUNDARY.md` — OCSF remains export-only

Audit ledger: docs/audits/AUDIT-2026-07-03.md (item T)

msaad00:
#3576 merged on main (`f007315ed`): multilang symbol reach (npm + Go), CWE/CVE/CPE advisory context, OSV/GHSA `affected_symbols` at build time.

Symbol join ecosystems: **pypi, npm, go**. `function_reachable` still requires advisory symbol data (sparse upstream).

Next: fold `symbol_reachability` into `effective_reach` scoring + VEX suggestions for `unreachable`.

msaad00:
Java (Maven) + Rust (Cargo) symbol join in #3579: AST `dependency_symbol_reach` for `.java`/`.rs` MCP entrypoints, wired into `apply_symbol_reachability_to_blast_radii`.

Merge queue: #3577 triage/VEX → #3579 multilang expansion → #3578 SARIF Finding.

msaad00:
## Reachability matrix update (post #3585, 2026-07-06)

| Layer | PyPI/npm/Go | Java/Rust | NuGet/C# | Ruby | Gradle-only |
|-------|-------------|-----------|----------|------|-------------|
| SBOM/CVE scan | ✅ | ✅ | ✅ | ✅ | ✅ |
| Graph reach | ✅ | ✅ | ✅ | ✅ | ✅ |
| `dependency_symbol` | ✅ | 🟡 regex | 🟡 regex | 🟡 regex | 🟡 regex |
| `symbol_reachability` | ✅ | 🟡 | 🟡 | 🟡 | 🟡 |
| Triage/VEX auto | ✅ | ✅ | ✅ | ✅ | ✅ |

**Ingest (evidence parity):**
- SARIF → `detect_and_parse` + `--external-scan` depth: **shipped** (#3585)
- CSAF / CycloneDX VEX ingest: **next** (quick-win #2, `docs/ROADMAP_QUICK_WINS.md`)

🟡 = manifest + regex proof (conservative; not full AST)

[...]

PR #3648, opened 2026-07-07
Part of #3499 (Iceberg catalog export)

msaad00:
## What

Registers Parquet finding exports into an **Iceberg REST catalog** as versioned table snapshots — the data-lake tail of #3499 (audit item T). Today only flat `.parquet` files are written (`parquet_fmt.py`); this adds an opt-in path that creates the namespace + table, appends a data file, and commits a snapshot so lake consumers query one consistent, versioned table.

## Design: pyiceberg vs thin client

Uses **pyiceberg's `RestCatalog`** rather than a hand-rolled REST client. A real Iceberg append must write Avro manifest + manifest-list files and per-column statistics for every commit; pyiceberg already does this and speaks the REST protocol, and it depends on `pyarrow` which the `[lake]` extra already ships. A thin client that only POSTs to the catalog cannot produce valid snapshots without reimplementing the Avro manifest writer — too heavy and error-prone. The dependency is optional and imported lazily (mirrors `parquet_fmt._require_pyarrow`); nothing runs unless a catalog URL is configured.

## What's wired

- **`output/iceberg_catalog.py`** (new) — `IcebergCatalogConfig` (env + args), `register_findings()` (namespace/table create-if-not-exists → append → snapshot), `maybe_register_iceberg()` (no-op unless configured). Schema comes from the shared Parquet schema, so no drift.
- **`parquet_fmt.py`** — extracted `to_arrow_table()` so both the flat-file writer and the Iceberg path reuse the identical 27-col schema + rows.
- **CLI** — `--iceberg-catalog-url` / `--iceberg-namespace` / `--iceberg-table` (env: `AGENT_BOM_ICEBERG_*`), disabled by default. Registration runs best-effort right after the `.parquet` write and degrades to a warning if the catalog/deps are unavailable — the file is already on disk.
- **`pyiceberg>=0.7`** added to the optional `[lake]` extra.

Config (env-driven, disabled by default): `AGENT_BOM_ICEBERG_CATALOG_URL`, `_NAMESPACE` (default `agent_bom`), `_TABLE` (default `findings`), `_CREDENTIAL` (OAuth2 `id:secret`), `_TOKEN` (bearer), `_WAREHOUSE`.

## Tests

`tests/test_iceberg_catalog.py` — 12 tests: config/request construction, disabled-by-default no-op, deps-absent graceful error (monkeypatched `_require_pyiceberg`), and round-trip against an in-memory fake REST catalog (no network) asserting namespace/table creation, the 27-col schema handed to Iceberg, snapshot append, and idempotent re-append. Existing `test_parquet_export.py` stays green (14 passed total).

## Tradeoff to note

pyiceberg caps `rich<15`, so the lock moves **rich 15.0.0 → 14.3.4** and **cachetools 7 → 6.2.6** project-wide. Core only requires `rich>=13`; the app imports and the full parquet/iceberg suite pass on 14.3.4. Flagging for awareness since `rich` is central to the CLI — no pyiceberg release allows `rich>=15`, so this is unavoidable if pyiceberg lands in the lock.

## Deferred (issue tail)

ClickHouse findings-ingest wiring and Swift bare-call precision remain separate items on #3499.

### DIFF
--- a/docs/operations/ENV_VARS.md
+++ b/docs/operations/ENV_VARS.md
@@ -171,6 +171,14 @@ so they cannot regress silently, but they are not part of this reference.
 |---|---|---|---|
 | `AGENT_BOM_HUB_OBSERVATIONS_RETENTION_DAYS` | `int` | `365` | Age-based retention for the Postgres occurrence log (``hub_findings_current_observations``). Monthly RANGE partitions are detached and dropped once wholly past this window. ``<= 0`` disables rollover. SQLite and legacy unpartitioned Postgre |
 
+## Iceberg Catalog Export
+| Env var | Type | Default | Description |
+|---|---|---|---|
+| `AGENT_BOM_ICEBERG_CATALOG_URL` | `str` | `''` | Optional findings-lake side-write. Disabled unless a REST catalog URL is set. Credentials are deliberately not mirrored here; they stay env/KMS-only and are allowlisted as secret material in scripts/env_var_allowlist.txt. |
+| `AGENT_BOM_ICEBERG_NAMESPACE` | `str` | `'agent_bom'` | — |
+| `AGENT_BOM_ICEBERG_TABLE` | `str` | `'findings'` | — |
+| `AGENT_BOM_ICEBERG_WAREHOUSE` | `str` | `''` | — |
+
 ## Image Scanning
 | Env var | Type | Default | Description |
 |---|---|---|---|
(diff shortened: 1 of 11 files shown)


msaad00:
Review note (blocking-ish): adding pyiceberg to the `[lake]` extra forces a **project-wide rich downgrade 15.0.0 → 14.3.4** in uv.lock, because pyiceberg caps `rich<15`. rich is central to the core CLI UX, so an *optional* data-lake feature shouldn't constrain it.

Recommend before merge: move pyiceberg out of the default locked resolution — either its own `[iceberg]` extra that's not part of the standard `uv sync` set, or document it as a manual `pip install pyiceberg` alongside the lazy-import (which already degrades gracefully when absent). Goal: base install keeps rich>=15; only users who opt into Iceberg pull pyiceberg + its rich cap. Functionality is fine either way (tests pass on 14.3.4); this is about not letting an optional feature pin the core's rich version.

agent-bom (approved):

andres-linero (approved):

Issue #3665, opened 2026-07-07
security(db): tenant RLS silently void when app connects as a Postgres superuser (default Docker role)


msaad00:
## Summary

Row-Level Security is the mechanism enforcing tenant isolation on the Postgres store (`api/postgres_common.py:155` sets `FORCE ROW LEVEL SECURITY` and the RLS policy keys on `current_setting('app.tenant_id')`). **Postgres superusers and `BYPASSRLS` roles ignore `FORCE ROW LEVEL SECURITY`** — so when the app connects as such a role, tenant isolation is silently and completely disabled.

This is not theoretical: the default Docker/quickstart connects as the role `agent_bom`, which is created as a **superuser**. Verified at scale — as the superuser role, a cross-tenant query (`SET app.tenant_id='benchB'; SELECT … WHERE scan_id='scan1'`) returned all 498,789 rows of the *other* tenant, not benchB's 3,784.

Isolation itself is correct: as a **non-superuser** role the same query correctly returns only the tenant's own 3,784 rows, and real app queries ride the `(tenant_id, scan_id)` composite index touching only tenant rows. The defect is purely that nothing stops the app from running as a role for which RLS is a no-op.

## Impact

- Any deployment using the default bundled Postgres (or any DB URL pointing at a superuser/`BYPASSRLS` role) has **no tenant isolation at the DB layer**, regardless of the RLS policies being present and correct. A bug or missing `WHERE tenant_id` in any query path leaks cross-tenant.
- Silent: startup succeeds, RLS policies show as enabled, tests on a non-superuser role pass — the gap only appears under the exact role the default ships with.

## Fix direction

1. Provision the app's DB role as `NOSUPERUSER NOBYPASSRLS` in the bundled Postgres init and in the Helm/compose defaults.
2. **Fail startup** if the connected role is a superuser or has `BYPASSRLS` (query `pg_has_role`/`rolsuper`/`rolbypassrls`) when multi-tenancy is enabled — loud, not silent.
3. Document the required role grants for customer-managed Postgres.

## Environment

Postgres 16.13, verified live against the 500k-node estate during graph-scale benchmarking.

msaad00:
**Validated against `main` — closed as shipped (via #3661).**

[...]

PR #3661, opened 2026-07-07
Closes #1522 (split output/html.py)

msaad00:
## What

Splits the last remaining #1522 monolith, \`src/agent_bom/output/html.py\` (3306 LOC), into a cohesive \`src/agent_bom/output/html/\` package. Pure mechanical refactor — **zero behaviour change**, public API preserved via package re-exports.

The other #1522 targets (postgres_store, ast_analyzer, mcp_server, output/__init__) were already resolved; this closes the issue.

## New layout

| Module | LOC | Responsibility |
|---|---|---|
| \`__init__.py\` | 55 | Re-exports public + semi-public symbols |
| \`_common.py\` | 29 | Colour constants, \`_esc\`, \`_sev_badge\` |
| \`sections.py\` | 1328 | Section renderers (summary/tables/graph-data/CIS/compliance) |
| \`styles.py\` | 234 | Embedded CSS/theme (\`render_styles\`) |
| \`scripts.py\` | 1363 | Embedded JS (tabs/pagination/graph) + offline post-processing |
| \`tabs.py\` | 55 | Tab-bar post-processing |
| \`document.py\` | 344 | Top-level assembler + file exporter (\`to_html\`, \`export_html\`) |

## Public API preserved

Every importer of \`agent_bom.output.html\` keeps working — the package re-exports \`to_html\`, \`export_html\`, \`_vuln_table\`, \`_cis_benchmark_section\`, \`_PAGE_SIZE\`, and the rest of the original module-level symbols. \`output/__init__.py\` re-exports are unchanged.

## Acceptance

- **Byte-for-byte identical HTML** verified against a fixture report (vulns + policy + CIS findings), both online and offline modes — \`diff\` empty.
- \`ruff check\` + \`mypy\` clean. The per-file \`E501\`/\`F541\` ignore was moved to the new package path; the \`f\` prefix on the static script string is load-bearing (it unescapes doubled braces), hence \`F541\` is ignored there.
- 338 HTML-touching tests pass (test_output_formats, test_html_report_scale, test_report_ux, test_core, test_cis_remediation_surfaces, test_exploit_likelihood, test_match_confidence_tier_exports, test_exposure_path_finding).

Closes #1522

PR #3667, opened 2026-07-07
security(db): fail startup when app DB role bypasses RLS (Part of #3665)

msaad00:
## The defect

Tenant isolation on Postgres is enforced entirely through the
`FORCE ROW LEVEL SECURITY` policies created in
`src/agent_bom/api/postgres_common.py:_ensure_tenant_rls` (the RLS policy keys
on `current_setting('app.tenant_id')`). **Postgres superusers and roles with
`BYPASSRLS` ignore `FORCE ROW LEVEL SECURITY`** — so when the app connects as
such a role, every `*_tenant_isolation` policy becomes a silent no-op and
cross-tenant reads/writes succeed.

The default bundled quickstart connects as `POSTGRES_USER` (`agent_bom`), which
the `postgres:16-alpine` image creates as a **SUPERUSER**. Result: tenant
isolation was silently, completely disabled on the default deployment.

## Fix 1 — fail closed at startup

`src/agent_bom/api/postgres_common.py` — new `_guard_rls_capable_role()` called
from `_get_pool()` right after the pool is established. It runs
`SELECT rolsuper, rolbypassrls, rolname FROM pg_roles WHERE rolname = current_user`
once per pool and raises `RlsRolePrivilegeError` (new, ~line 45) when the role is
`SUPERUSER` or `BYPASSRLS`. The message names the role and the fix
(`ALTER ROLE <role> NOSUPERUSER NOBYPASSRLS` or point at `agent_bom_app`).

**Escape hatch:** `AGENT_BOM_ALLOW_SUPERUSER_DB=1`
(`src/agent_bom/config.py`, `ALLOW_SUPERUSER_DB`) downgrades the hard error to a
one-time warning for single-tenant / local dev. The probe is best-effort — a
transient connect error never masks the primary failure path.

Fail-closed by default; opt-in fail-open via the env var (documented in the
config docstring).

## Fix 2 — provision the bundled role correctly

`deploy/supabase/postgres/init.sql`:
- `agent_bom_app` is now created `NOSUPERUSER NOBYPASSRLS` explicitly.
- A final init step strips `SUPERUSER`/`BYPASSRLS` from the schema-owning
  `agent_bom` role (`ALTER ROLE ... NOSUPERUSER NOBYPASSRLS`). It runs last,
  after all extensions/tables are created; ownership is preserved so the app's
  runtime idempotent DDL still works, but the role is now subject to its own
  tenant RLS policies. The attribute change only applies on the next connection,
  so the rest of init is unaffected.

DDL is not broken: a non-superuser **owner** can still `ALTER TABLE`,
`ENABLE`/`FORCE` RLS, and `CREATE POLICY`/`FUNCTION`, and `FORCE ROW LEVEL
SECURITY` correctly applies to a non-superuser owner. No migration/runtime role
split was required.

## Tests

`tests/test_rls_superuser_guard.py`: guard raises for SUPERUSER and for
BYPASSRLS; escape hatch downgrades to a warning; non-privileged role passes;
one-time semantics; best-effort on probe failure.

```
python -m pytest tests/ -k "tenant or rls or postgres or isolation" -p no:randomly -q
# 533 passed, 6 skipped
```

Part of #3665

### DIFF
--- a/deploy/supabase/postgres/init.sql
+++ b/deploy/supabase/postgres/init.sql
@@ -1135,11 +1135,15 @@ BEGIN
     app_pass := current_setting('init.app_password', true);
 
     IF app_pass IS NOT NULL AND app_pass != '' THEN
-        -- Create app user if not exists
+        -- Create app user if not exists.
+        -- NOSUPERUSER NOBYPASSRLS is explicit (and the CREATE ROLE default) so
+        -- FORCE ROW LEVEL SECURITY tenant policies are always enforced for the
+        -- app role. A superuser/BYPASSRLS role silently voids tenant isolation
+        -- (#3665).
         IF NOT EXISTS (SELECT 1 FROM pg_roles WHERE rolname = 'agent_bom_app') THEN
-            EXECUTE format('CREATE ROLE agent_bom_app LOGIN PASSWORD %L', app_pass);
+            EXECUTE format('CREATE ROLE agent_bom_app LOGIN NOSUPERUSER NOBYPASSRLS PASSWORD %L', app_pass);
         ELSE
-            EXECUTE format('ALTER ROLE agent_bom_app PASSWORD %L', app_pass);
+            EXECUTE format('ALTER ROLE agent_bom_app NOSUPERUSER NOBYPASSRLS PASSWORD %L', app_pass);
         END IF;
 
         -- Connection limit: prevent app from exhausting all connections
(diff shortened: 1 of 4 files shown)


msaad00:
Superseded by #3661. I folded the RLS startup guard scope into the combined HTML/dependency/security PR, regenerated docs/operations/ENV_VARS.md, and added the RBAC policy.manage fix that was causing the Alpine approval-queue failure.

msaad00:
Closing as superseded by #3661, which now carries this scope.

msaad00:
Regression review = correctly wired and fail-closed (single global pool via `_get_pool()`, guard on the real path, best-effort probe retries rather than permanently skipping, `WHERE rolname=current_user` is the right effective-privilege check). Three refinements to fold in before/at merge:
1. **Escape hatch shouldn't silently cover multi-tenant.** `AGENT_BOM_ALLOW_SUPERUSER_DB=1` downgrades to one log line and doesn't distinguish single- from multi-tenant. Because the bundled quickstart connects as the superuser by default, **every existing default deployment trips the guard on upgrade** → operator pressure to just set the flag → a genuinely multi-tenant operator who does so silently voids all isolation. Refuse the escape hatch when multi-tenancy is actually configured, and surface the warning on an operator/health surface, not only logs.
2. **Also change the bundled compose default to the non-superuser `agent_bom_app` role** (not just strip in init.sql) so operators aren't nudged toward the flag in the first place.
3. **init.sql role-strip only helps fresh installs** (runs once on an empty data dir). Existing upgrades keep the superuser `agent_bom` role — add an explicit migration note (`ALTER ROLE ... NOSUPERUSER NOBYPASSRLS`).
Also: dropping SUPERUSER means runtime `CREATE EXTENSION IF NOT EXISTS pg_trgm` (`postgres_graph.py:371`) needs pg_trgm pre-created (bundled init does this) or the role to retain DB `CREATE` on PG<13 — worth a doc note for non-bundled Postgres.

Issue #3664, opened 2026-07-07
perf(graph): O(V+E)-per-request materialization — graph endpoints degrade to 14–170s at 500k nodes


msaad00:
## Summary

Benchmarking the graph subsystem at scale (493,221 nodes / 1,363,200 edges, Postgres 16) shows a **systemic O(V+E)-per-request bottleneck**: every endpoint backed by `PostgresGraphStore.load_graph()` re-materializes the *entire* tenant graph into a fresh in-memory `UnifiedGraph` on **every request** — no cache, no snapshot reuse, no pushed-down traversal. The DB is not the problem (bitmap index scan returns 493k nodes in ~328 ms); the wall is Python rebuilding ~493k `UnifiedNode` + ~1.36M `UnifiedEdge` objects (plus per-node `json.loads`) per call.

Search and base keyset pagination stay sub-second and flat — those are the model to follow.

## Measured latency (500k nodes, median of 3, single-worker uvicorn + Postgres 16)

| Endpoint | 100k | 500k | Loads full graph? |
|---|--:|--:|---|
| `/v1/graph/search` (trigram index) | 16 ms | **40 ms** flat ✅ | No |
| `/v1/graph` paged (limit 500) | 0.46 s | 1.88 s | No, but embeds full edge COUNT |
| `/v1/graph?relationships=…` (filtered) | 2.56 s | **21.3 s** | Yes |
| `/v1/graph/rollup` | 2.31 s | **16.4 s** (173 MB body) | Yes |
| `/v1/graph/clusters` | 2.10 s | 17.0 s | Yes |
| `/v1/graph/impact` depth=1 | 1.72 s | 14.3 s | Yes (full graph for a 1-hop) |
| `/v1/graph/paths` (BFS depth 3) | 1.74 s | 13.1 s | Yes |
| `/v1/graph/governance` | 2.25 s | 18.7 s | Yes |
| `/v1/graph/nhi/governance` | 28.2 s | **169.5 s** | Yes (worst) |
| `/v1/graph/diff` (500k vs 500k) | — | **73.8 s** | Loads all nodes+edges of BOTH scans |

At scale, normal traffic exceeds the `adaptive_backpressure` graph threshold (p99 > 2500 ms), so the safety valve 429-sheds the *entire* graph path for ~12 s windows — a self-inflicted availability outage that is really a symptom of this bottleneck.

## Ranked defects

[...]

PR #3669, opened 2026-07-07
perf(graph): bounded snapshot-stats, rollup orphans, deep-offset cap (Part of #3664)

msaad00:
Three bounded, high-confidence graph-scalability fixes from #3664. Excludes the `load_graph()` cache/CTE rearchitecture (P0 #1/#2), which needs separate design.

## Fix 1 — `snapshot_stats` re-counted edges on every paged read (~3s at 500k)
`snapshot_stats` ran `COUNT(*)` over `graph_edges` with a double `source_id IN (SELECT … graph_nodes) AND target_id IN (SELECT …)` subquery on **every** paged `/v1/graph` call.

- `src/agent_bom/api/postgres_graph.py` `snapshot_stats` (~line 1583)
- `src/agent_bom/api/graph_store.py` `snapshot_stats` (SQLite, ~line 1291)

The node/edge totals are already materialised on the `graph_snapshots` row (`node_count`, `edge_count`) at write time. The unfiltered path now reads those stored counts and only falls back to `COUNT(*)` when the snapshot row is missing or the counts were never populated. Any active entity-type or severity filter still recomputes (correctness preserved).

## Fix 2 — `/v1/graph/rollup` returned an unbounded body (173 MB at 500k)
Every non-CONTAINS node became its own `top_level` orphan entry, so the response grew O(nodes) on a flat estate.

- `src/agent_bom/graph/rollup.py` `rollup_view` (~line 346-366)

The highest-risk orphans are surfaced individually up to a cap (200); the remaining tail is aggregated into an additive `orphan_summary` (by-type counts, severity histogram, exposure/toxic rollup, and a bounded per-item sample). The cloud CONTAINS rollup tree is untouched. Existing response keys are preserved; `orphan_count` keeps its meaning (total orphans), with `orphan_shown_count` / `orphan_truncated_count` and `orphan_summary` added. UI types in `ui/lib/api-types.ts` updated additively.

## Fix 3 — deep OFFSET pagination is O(offset) (`offset=490000` = 6.2s)
Keyset `cursor=` is the flat path; deep offsets still force the store to walk and discard every skipped row.

- `src/agent_bom/api/postgres_graph.py` / `graph_store.py` `page_nodes` OFFSET branch
- `src/agent_bom/api/routes/graph.py` (`get_graph`, `list_graph_agents`)

Offsets beyond `MAX_NODE_PAGE_OFFSET` (10,000) without a cursor are rejected with a 422 whose message points callers to `cursor=`. Offset paging below the cap and all cursor paging are unchanged. Guarded at the route layer, with a store-level `ValueError` backstop for direct callers.

## Measured before-numbers (from #3664)
- snapshot-stats edge recount: ~3s per paged call at 500k
- rollup body: 173 MB at 500k
- `offset=490000`: 6.2s

## Tests
- New `tests/test_graph_scale_quickwins.py` (9 cases): stored-count read + NULL fallback + filtered recompute; offset cap at store and route (`/v1/graph`, `/v1/graph/agents`); orphan bounding, tail aggregation, highest-severity-orphan retention, and small-estate no-op.
- `pytest tests/ -k graph` green (714 passed, 3 skipped); rollup/graph-api suites green.

Part of #3664


msaad00:
Regression review = SAFE. Stored-count fallback is correct (trusted only when `not filters_active and value is not None`, else `COUNT(*)`; write path refreshes both counts atomically in the same txn as node/edge inserts; retention deletes the snapshot row with its rows, so counts can't outlive their data). Offset cap guards both `page_nodes` callers. Two consumer-visible notes for sign-off, non-blocking:
1. **`top_level` now capped at 200** (was every orphan). Additive keys aside, a consumer treating `top_level` as the complete node list silently loses orphans beyond 200 (only 20 appear in `orphan_summary.sample`). Confirm the UI reads `orphan_summary` for the full picture. (This is the PR's intent — bounding the 170MB body — just flagging the content change.)
2. **Edge-count semantics shift:** stored `edge_count = len(graph.edges)` counts all saved edges; the replaced COUNT used a both-endpoints-in-graph_nodes membership subquery. If the builder ever persists dangling edges, reported `total_edges` now exceeds pageable edges (arguably more correct). Low impact.

andres-linero (approved):

agent-bom (approved):

PR #3672, opened 2026-07-07
refactor(cli): unify scan verb, tier flags behind --help-all, dedup formats

msaad00:
## Why

The CLI front door was inconsistent: the tagline sells `connect → scan → graph → report`, but `scan` was a hidden alias while `agents` was the visible verb; `scan --help` dumped ~200 flat options; two formats were silent duplicates; and three verbs (`serve`/`api`/`up`) all start a server. This PR tightens the surface while preserving every existing invocation.

Every finding was verified in code before changing it. All flags remain functional — only help visibility/grouping changed.

### FIX 1 — Unify the scan verb
`cli/__init__.py` — `scan` is now the visible canonical verb; `agents` is a hidden back-compat alias (was reversed).
- Before: `main.add_command(_agents_cmd, "agents")` visible; hidden `scan` clone.
- After: `main.add_command(_agents_cmd, "scan")` visible; hidden `agents` clone.
- Banner (`_print_startup_banner`) + top-level docstring now say `agent-bom scan`; `_grouped_help.py` "Get started" lists `scan`.
- `agent-bom agents ...` still works (hidden).

### FIX 2 — Tier the flags behind `--help-all`
New `cli/_scan_help.py::TieredCommand` (applied via `@click.command(cls=TieredCommand)` on `scan`).
- `scan --help` now renders **Core options** + **More options**; ~33 third-party credential/endpoint flags (`--jira-*`, `--slack-webhook/-bot-token`, `--servicenow-*`, `--snyk-token/-org`, `--drata-token`, `--vanta-token`, `--wandb-*`, `--nebius-*`, `--openai-api-key/-org-id`, `--hf-token/-username/-organization`, `--iceberg-*`, `--siem-*`, `--push-url/-api-key`, `--clickhouse-url`) are hidden until `scan --help-all`, each shown with its backing env var.

### FIX 3 — Dedup output formats
`cli/options_sources.py` + new `AliasedChoice`.
- `text` dropped from the advertised `-f/--format` choices; still accepted, maps to `plain`.
- `graph-html` **retained** — verified it renders a distinct interactive graph view via `export_graph_html` on file write (the shared `to_html()` is only an unreachable stdout fallback), so removing it would have been a regression.

### FIX 4 — Collapse the server triplet
`cli/_server.py` + `api/server.py`.
- New `serve --no-ui` provides REST-only mode (sets `AGENT_BOM_NO_UI`, which gates `_mount_dashboard`).
[...]

PR #3673, opened 2026-07-07
test: eliminate order-dependent flake class (global-singleton isolation + enable pytest-randomly)

msaad00:
## Problem

Tests that swapped a process-global backend via one of the ~35 `set_*_store` accessors (`api/stores.py` plus 15 sibling store modules) without restoring it leaked that singleton into later tests on the same worker — making outcomes depend on collection order. Two such flakes already reached `main` (NO_AUTH_ROLE poisoning #3660, leaked durable job store behind the MCP/HTTP shape flake #3663). The autouse conftest reset covered only a handful of stores, and `pytest-randomly` was installed but disabled with a stale "suite is order-independent" claim.

## Fix (test isolation only — no `src/` behavior change)

**Snapshot/restore every global singleton per test** in `tests/conftest.py`:
- Discovers store singletons by **introspecting each store module** (any global whose name ends in `_store`, plus a few explicit extras) rather than a hand-maintained list — new stores are covered automatically. Functions/classes are filtered out; only state holders are captured, then restored verbatim on teardown.
- Snapshot/restore the shared Rich output console (`agent_bom.output.console`) — a test that swaps it for a capture console and skips its restore made later CLI tests render into the stale console (empty `result.output`).
- Reset the proxy route in-memory buffers (`_proxy_alerts` / `_proxy_alerts_total` / `_proxy_metrics` / `_proxy_metrics_by_tenant`) — `_load_proxy_alerts` only falls back to the audit log when `_proxy_alerts` is empty, so a leaked alert reported the wrong `/v1/proxy/status` count.
- Snapshot/restore the auth-posture config constants as a backstop.

**Randomization is now ON by default:** dropped `-p no:randomly` from pyproject `addopts` and the CI test step. A failing run stays reproducible — pytest-randomly prints `Using --randomly-seed=N`; re-run with `--randomly-seed=N` to replay the exact order.

## Flakes found & fixed by turning randomization on

| Test | Root cause |
|---|---|
| `test_api_proxy_scorecard.py::test_proxy_status_from_log` | leaked `_proxy_alerts` deque suppressed the audit-log fallback → `total_alerts` 0 vs 1 |
| `test_cli_history.py::test_diff_quiet_prints_compact_summary` | leaked capture `agent_bom.output.console` → empty CLI output |

## Verification (all green)

- **Subset** (`test_api_*`, `test_mcp_*`, `test_cli_*`, `test_hitl_*`, `test_hardening_*` — 1761 tests): randomly seeds **1, 2, 3** and **reverse collection order** → 1760 passed, 1 skipped each.
- **Full suite** (12,947 tests, `-m "not network and not slow"`): serial randomly seeds **4, 5, 6** → 0 failures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -693,14 +693,15 @@ jobs:
         run: |
           # -n auto --dist loadscope: spread tests across worker processes,
           # grouping same-module tests on one worker so module-scoped fixtures
-          # behave. -p no:randomly keeps CI in deterministic order (reproducible
-          # failures); it is redundant with the pytest addopts default but kept
-          # explicit. The suite is order-independent — run `pytest -p randomly`
-          # locally to verify.
+          # behave. Test order is randomized by default (pytest-randomly, ON via
+          # pyproject addopts) to keep order-dependence from regressing; the seed
+          # is printed as "Using --randomly-seed=N" and a failure replays exactly
+          # with `pytest --randomly-seed=N`. Isolation is guaranteed by the
+          # global-singleton snapshot/restore fixture in tests/conftest.py.
           if [ "${{ matrix.python-version }}" = "3.11" ] && [ "${{ github.event_name }}" != "pull_request" ]; then
-            uv run pytest tests/ -v --tb=short --cov=agent_bom --cov-report=term-missing --cov-fail-under=75 -m "not network" -n auto --dist loadscope -p no:randomly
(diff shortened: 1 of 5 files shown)


msaad00:
Superseded by #3672. The global-singleton isolation and pytest-randomly scope is now folded into the combined CLI/UI/test PR.

msaad00:
Closing as superseded by #3672, which now carries this scope.

PR #3852, opened 2026-07-12
docs(readme): add read-only live demo link to the hero row

msaad00:
Adds a hero **Live demo** link to the always-on anonymous sandbox at demo.agent-bom.com (read-only estate, verified healthy), next to Docs. Labeled 'read-only sandbox' so it reads as a try-it demo, not managed SaaS; Self-host stays in the row as the real path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@
 <p align="center">Headless agent primitives and human cockpit surfaces over one shared evidence model.</p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


msaad00:
Superseded by #3855, which combines this README demo link with the provider onboarding matrix and Dependabot SDK grouping into one verified release-follow-up PR.

PR #3864, opened 2026-07-12
fix(security): confine POST /v1/scan local paths to the same jail as secondary scan endpoints

msaad00:
## The gap

The primary `POST /v1/scan` flow (`create_scan` → `enqueue_scan_job` → pipeline) only ran `validate_path(p, must_exist=True)` on its local-path fields. With the default `restrict_to_home=False`, that guard rejects `..` traversal but **accepts absolute paths and does not confine to a configured root**. The dedicated scan endpoints (`/scan/dataset-cards`, `/scan/prompt-scan`, etc.) route their paths through the hardened `_sanitize_api_path` / `_api_scan_path_or_400` helper (gated by `AGENT_BOM_API_LOCAL_PATH_SCANS`, default disabled) — but the primary endpoint bypassed it entirely.

Result: an authenticated caller could `POST /v1/scan {"inventory":"/etc/hosts"}` (or `filesystem_paths`, `sbom`, `tf_dirs`, `gha_path`, `jupyter_dirs`, `agent_projects`) and the control plane would read arbitrary host files, with contents partly reflected via parse errors / results. `external_scan` and `vex` were opened with **no** validation at all.

### Repro (before the fix)

```
validate_path('/etc/hosts', must_exist=True)  -> /etc/hosts        (no rejection)
POST /v1/scan {"inventory":"/etc/hosts"}       -> 202, inventory reader receives /etc/hosts
```

## The fix

Route **every** `ScanRequest` local-path field through the existing `_api_scan_path_or_400` helper and the same `_api_local_scans_enabled()` gate the secondary endpoints use, at request ingest in `create_scan` — before the job is queued. Fields covered: `filesystem_paths`, `agent_projects`, `tf_dirs`, `gha_path`, `inventory`, `jupyter_dirs`, `sbom`, and now `external_scan` and `vex` (previously opened unvalidated).

With the gate at its default (disabled) posture, local-path scans are rejected with a clean `400 Invalid scan path` on the primary endpoint too — consistent with the secondary endpoints. Image / connector / k8s / `repo_url` scans are unaffected. When the gate is enabled and paths resolve under the configured scan root, legitimate local scans still run.

The helper is reused unchanged — not reimplemented or weakened.

## Bundled hardening

- **Backend model tightening** (`ScanRequest`): `format` and `min_severity` are now enum-constrained (with a case/whitespace-normalizing pre-validator so `HIGH`/`CycloneDX` still work) — junk is rejected at the edge with `422` instead of flowing downstream.
- **Frontend defense-in-depth** (`ui/components/scan-form.tsx`): the public-repo URL field is client-side validated as a well-formed `http(s)://<host>` before submit is enabled, with an inline error; path inputs are trimmed. (`findings-lens.ts` intentionally untouched.)

## Tests

- New endpoint tests assert absolute and `..` paths to `POST /v1/scan` are rejected (400) with the gate at default, across all nine path fields incl. `external_scan`/`vex`; that non-path (image/k8s/repo_url) scans still succeed; and that a valid relative path under the root scans when the gate is enabled.
- New model tests assert `422` on bad `format`/`min_severity` and normalization of case variants.
- New UI test asserts an invalid repo URL disables submit and shows the inline error.
- Full `tests/api/` suite (698) green; security suites (120) green; scan/type/batch/sources (100) green; `ui` typecheck clean; scan-form vitest (7) green.

### DIFF
--- a/docs/openapi/v1.json
+++ b/docs/openapi/v1.json
@@ -2895,6 +2895,14 @@
           },
           "format": {
             "default": "json",
+            "enum": [
+              "json",
+              "cyclonedx",
+              "sarif",
+              "spdx",
+              "html",
+              "text"
+            ],
             "title": "Format",
             "type": "string"
           },
@@ -2957,6 +2965,12 @@
           "min_severity": {
             "anyOf": [
               {
+                "enum": [
+                  "low",
+                  "medium",
+                  "high",
+                  "critical"
+                ],
                 "type": "string"
               },
               {
--- a/docs/openapi/v1.yaml
+++ b/docs/openapi/v1.yaml
@@ -1915,6 +1915,13 @@ components:
           type: array
         format:
           default: json
+          enum:
+          - json
+          - cyclonedx
+          - sarif
+          - spdx
+          - html
+          - text
           title: Format
           type: string
         gha_path:
@@ -1952,7 +1959,12 @@ components:
           title: K8S Namespace
         min_severity:
           anyOf:
-          - type: string
+          - enum:
+            - low
+            - medium
+            - high
+            - critical
+            type: string
           - type: 'null'
           title: Min Severity
         no_scan:
--- a/docs/schemas/v1/ScanJob.json
+++ b/docs/schemas/v1/ScanJob.json
@@ -107,6 +107,14 @@
         },
         "format": {
           "default": "json",
+          "enum": [
+            "json",
+            "cyclonedx",
+            "sarif",
+            "spdx",
+            "html",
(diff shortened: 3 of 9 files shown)


msaad00:
Temporarily closing and reopening to retrigger missing required pull-request checks; no code or branch history changes.

agent-bom (dismissed):

andres-linero (dismissed):

andres-linero (approved):

agent-bom (approved):

PR #3877, opened 2026-07-13
fix(inventory): facet-filtered list drops rows past the page limit (P1, pre-tag)

msaad00:
Pre-release bug-hunt found a **P1** in the new unified inventory (headline 0.95.0 feature): the facet-filtered `/v1/inventory/assets` list — and the `inventory_list` MCP tool — **silently drop matching rows**.

**Cause:** the continuation cursor was set to the store cursor after the last *consumed* store page, so facet matches collected beyond `limit` within that final store page were skipped, and `has_more` reported `False`. Repro: 400 matches, limit 50 → 300 dropped, page reported complete.

**Fix:** when more matches than `limit` are collected, mint the continuation cursor from the **last emitted node** (`encode_graph_cursor`, the same encoder the store uses), so the next request resumes exactly after it and every match is returned once. Adds a multi-page facet regression test (130 matches, limit 40) — **verified it fails on the old logic, passes on the fix**. Ruff clean; full inventory + MCP inventory suite green (24).

### DIFF
--- a/src/agent_bom/api/inventory_service.py
+++ b/src/agent_bom/api/inventory_service.py
@@ -19,7 +19,7 @@ from __future__ import annotations
 import asyncio
 from typing import Any, Awaitable, Callable, Optional
 
-from agent_bom.api.graph_store import MAX_NODE_PAGE_OFFSET
+from agent_bom.api.graph_store import MAX_NODE_PAGE_OFFSET, encode_graph_cursor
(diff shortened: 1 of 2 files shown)


msaad00:
Temporarily closing and reopening to retrigger missing required pull-request checks; no code or branch history changes.

agent-bom (approved):

andres-linero (approved):

PR #3960, opened 2026-07-14
feat(deploy): keyless out-of-box control-plane cloud identity + cross-account connect

msaad00:
## Summary

Make the control plane's **cloud identity** provisioned and documented so a real
deployment connects clouds with **zero manual base identity and zero static
keys**. The control plane reads a customer cloud by *assuming* that account's
read-only role (`sts:AssumeRole` + ExternalId) — it never stores customer keys —
so it only needs its own keyless workload identity to make that call.

**Gap found and fixed:** the scanner IRSA role minted by `aws/baseline` attached
only the direct-read policy, which has **no `sts:AssumeRole`**. So the keyless
control-plane identity could read only the account it runs in; AWS Organizations
fan-out and hosted connect (assume a read-only role in each target account) both
failed. This PR grants a scoped `sts:AssumeRole`.

### Terraform (the load-bearing change)
- `aws/baseline`: attach a **least-privilege `sts:AssumeRole`** policy to the
  scanner IRSA role, scoped to the connection-role name pattern
  (`agent-bom-readonly*` / `abom-readonly*`) via a new `connect_role_arns`
  variable (default on; `[]` disables). Only `sts:AssumeRole` is granted — no
  broader action, never admin. The connection role's own trust policy +
  ExternalId remain the reciprocal guard. New output
  `scanner_assume_connect_policy_arn`.
- `platform-eks`: pass `connect_role_arns` through to baseline, expose the
  output, document in README.

### Non-EKS keyless identity paths (documented + configurable)
- EKS → scanner IRSA role • self-managed k8s on EC2/ECS → node instance profile
  / ECS task role (AWS SDK default chain) • AKS → Azure workload identity • GKE →
  Workload Identity. Documented in `values.yaml`, a new
  `examples/control-plane-identity-values.yaml`, and the docs. No static keys in
  any path (Snowflake key-pair via Secret reference is the sole non-workload-identity exception).

### Docs
- `DEPLOY_PLATFORM.md`: guided **"Connect your cloud (zero keys)"** section —
  deploy (identity auto-provisioned) → read the identity ARN per target → create
  the read-only role in each account trusting that identity + ExternalId, or run
  the org **StackSet** for all accounts at once → connect in the UI. Cross-links
  the CloudFormation StackSet and `connect-*` modules.
- `SECURITY_ARCHITECTURE.md`: **"Control-plane identity"** section — workload
  identity per target, assumes connection roles, ExternalId, least-privilege, no
  static keys, and the laptop/dev path is explicitly dev-only.

## Verification
- `terraform validate` + `fmt -check -recursive`: **clean** on `aws/baseline`
  and `platform-eks` (one pre-existing `aws_region.name` deprecation warning,
  unrelated).
- `helm lint` + `helm template` with `control-plane-identity-values.yaml`:
  **clean**; the keyless role-arn annotation renders on the control-plane and
  scanner ServiceAccounts.
- Internal doc links + anchors resolve.
- No secrets in any file; no Python/contract changes.

## Notes
- The connect **wizard** "trust the control-plane identity" UI change (surfacing
  the identity ARN to paste into the trust policy) is a **separate follow-up**,
  per the onboarding epic — this PR is deploy configs + docs only.
- Cross-account assume is default-on with the documented name pattern for
  out-of-box "just works"; operators can narrow `connect_role_arns` to explicit
  ARNs or disable it.

### DIFF
--- a/deploy/helm/agent-bom/examples/README.md
+++ b/deploy/helm/agent-bom/examples/README.md
@@ -20,6 +20,7 @@ product split.
 | `oidc-discovery-shim` | `eks-mcp-pilot-values.yaml` + `gateway-upstreams.example.yaml` + `oidc-discovery-shim-values.yaml` | Gateway-hosted `/.well-known/openid-configuration` for legacy IdPs (MCP OAuth interop) |
 | `snowflake-backend` | `eks-snowflake-values.yaml` | Overlay for Snowflake governance and selected store parity, not a claim of full control-plane replacement |
 | `gateway-runtime` | `eks-mcp-pilot-values.yaml` + `gateway-upstreams.example.yaml` | Focused pilot plus central gateway rendering for shared MCP relay/policy |
+| `control-plane-identity` | `control-plane-identity-values.yaml` | Keyless control-plane cloud identity (EKS IRSA / EC2-ECS instance role / AKS + GKE workload identity) that assumes read-only connect roles — zero static keys. See `docs/DEPLOY_PLATFORM.md` "Connect your cloud (zero keys)" |
 
 ## Validate the shipped profiles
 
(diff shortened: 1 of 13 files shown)


msaad00:
Superseded by #3970, which now preserves this PR's signed feature commit and combines it with the two requested Node/toolchain updates. The exact combined tree was revalidated with Terraform, Helm, and UI checks.

PR #3973, opened 2026-07-14
fix(graph): honor AGENT_BOM_STATE_DIR and stale-aware demo seed

### LINKED ISSUE #3964: graph: honor AGENT_BOM_STATE_DIR + stale-aware showcase seeding (stale/seed snapshots pollute the graph)
Validated vs main @ eb58d36bb (audit P1-5, CONFIRMED) — also live-reproduced: after a real AWS scan the Security Graph showed 25 accumulated snapshots + a 7/13 demo seed, not the fresh scan. `default_graph_db_path` honors only AGENT_BOM_GRAPH_DB/AGENT_BOM_DB, never AGENT_BOM_STATE_DIR (graph_store.py:251-263; contrast durable_store.py:48). `seed_showcase_graph_if_empty` early-returns on ANY existing snapshot — seed-if-empty, not seed-if-stale (showcase_graph.py:594-597). Fix: honor the state dir; stale-aware seeding; default the graph to the latest scan.

msaad00:
Closes #3964.

## Problem
After a real AWS scan the Security Graph showed 25 accumulated snapshots plus a stale demo seed instead of the fresh scan, and the local `~/.agent-bom/db/graph.db` had grown to hundreds of MB of test/demo pollution. Two root causes:

1. **State dir ignored.** `default_graph_db_path` honored only `AGENT_BOM_GRAPH_DB` / `AGENT_BOM_DB`, never `AGENT_BOM_STATE_DIR` — unlike `api.durable_store`. So every run and the whole test suite wrote the real `~/.agent-bom/db/graph.db` even when state was redirected off `$HOME` (the conftest sets `AGENT_BOM_STATE_DIR` but not `AGENT_BOM_GRAPH_DB`), accumulating cross-run/test snapshots.
2. **Seed-if-empty, not seed-if-stale.** `seed_showcase_graph_if_empty` early-returned on *any* existing snapshot, so a polluted DB booted `--demo-estate` on stale seed data.

## Fix
- `default_graph_db_path` now honors `AGENT_BOM_STATE_DIR` (`<state>/db/graph.db`), keeping the `AGENT_BOM_GRAPH_DB` / `AGENT_BOM_DB` overrides ahead of it — matching `durable_store`'s precedence. This also stops the test suite from polluting the real home graph DB.
- `seed_showcase_graph_if_empty` is now stale-aware:
  - a real (non-showcase) scan is **never** shadowed (skip seeding),
  - a current demo seed is a no-op,
  - a stale demo seed (missing baseline or out-of-date `created_at`) is cleared and re-seeded.
- Note on "default to latest scan": the read path already selects the newest snapshot by `created_at DESC`, and the showcase seed is pinned to fixed past timestamps, so a fresh scan (today) is naturally the one served — no separate change needed here.

## Tests (TDD)
- `tests/test_graph_store_state_dir.py` — state-dir precedence + the real home dir is untouched when a state dir is configured.
- `tests/test_showcase_graph_seeding.py` — seeds when empty; idempotent when current; re-seeds when stale; does **not** shadow a real scan.

## Verification
- New suites pass; existing `test_demo_estate_bootstrap.py` (24) + `test_graph_store_mkdir.py` pass.
- `mypy` + `ruff` clean on changed files; `make preflight` green.

### DIFF
--- a/src/agent_bom/db/graph_store.py
+++ b/src/agent_bom/db/graph_store.py
@@ -255,11 +255,20 @@ def default_graph_db_path() -> Path:
     Preference order:
     1. ``AGENT_BOM_GRAPH_DB`` explicit graph database path
     2. ``AGENT_BOM_DB`` shared SQLite database used by the API
-    3. ``~/.agent-bom/db/graph.db`` local default
+    3. ``AGENT_BOM_STATE_DIR`` per-process state dir (``<state>/db/graph.db``)
+    4. ``~/.agent-bom/db/graph.db`` local default
+
+    Honoring ``AGENT_BOM_STATE_DIR`` mirrors ``api.durable_store`` so the graph
+    snapshot lands in the same isolated state dir an operator (or the test
+    suite's conftest) redirects state to — instead of always writing the real
+    ``~/.agent-bom/db/graph.db`` and accumulating cross-run/test pollution.
     """
     configured = os.environ.get("AGENT_BOM_GRAPH_DB") or os.environ.get("AGENT_BOM_DB")
     if configured:
         return Path(configured).expanduser()
+    state_dir = os.environ.get("AGENT_BOM_STATE_DIR")
+    if state_dir:
+        return Path(state_dir).expanduser() / "db" / "graph.db"
     return Path.home() / ".agent-bom" / "db" / "graph.db"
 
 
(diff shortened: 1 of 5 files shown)


msaad00:
Temporarily closing and reopening to retrigger missing pull_request checks; the head commit is unchanged.

andres-linero (approved):

agent-bom (approved):

PR #3974, opened 2026-07-14
fix(overview): honest exec read off the unified findings spine + offload

msaad00:
## Problem

The exec headline (critical/high) and the risk **grade** were computed from `_scan_rollup` → `_rollup_from_blast_radius` (CVE-only), while the coverage lanes and `/v1/findings` read the **unified findings spine** `result["findings"]`. The pipeline adds non-CVE findings beyond `blast_radius` (malicious/blocklisted package, MCP/a2a auth, IaC). So a scan-produced **critical malicious/blocklisted package was invisible in the exec headline critical/high and the grade** — the one correlated evidence layer broke on its own headline. Plus `compliance_failing` was hardcoded `0` (grade never moved on compliance), and `get_overview` ran all store/DB work inline on the event loop with an unindexed `json_extract` severity GROUP BY.

## What changed

**One reconciled rollup off the unified spine** (`_estate_rollup`): the headline severity, the five coverage lanes, the top-risk strip, and the grade are all built from a single pass over the same `findings` stream. `blast_radius` / compact `summary` remain **fallbacks** only when a completed scan carries no spine (e.g. post hot-cache compaction), so a non-CVE critical now reaches the headline + grade and can never be visible in one surface and invisible in another. Closes #3961.

**Live compliance into the grade** (Closes #3962): the count of failing compliance frameworks (a framework with a critical/high finding fails) is accumulated in the *same* rollup and fed to `compute_exec_score` — no second control-by-control evaluation on the hot endpoint.

**Offload + index** (Closes #3963): `get_overview` offloads the store/DB work off the event loop via `anyio.to_thread.run_sync` under the adaptive backpressure guard (sheds with `429` + `Retry-After`). SQLite + Postgres `severity_breakdown` now GROUP BY the materialized, indexed `severity` column instead of an unindexed per-row `json_extract` / payload JSON decode; a sargable `(tenant_id, LOWER(severity))` index is added to the Postgres ledger.

No API field renames — the UI reads `overview.headline.*` directly and needs no change.

## Tests (TDD — failing test written first for each)

- `test_overview_headline_reflects_noncve_spine_critical` — a spine-only critical (absent from blast_radius) reaches the headline + grade + coverage.
- `test_overview_compliance_failing_moves_grade` — mapping the same critical to failing frameworks drops the score via the compliance driver.
- `test_overview_sheds_with_429_when_backpressure_opens` — proves the offload wrapping.
[...]

PR #3979, opened 2026-07-14
fix(cloud): Snowflake CIS benchmark over REST + guard in-memory hub sort ceiling

### LINKED ISSUE #3967: backlog: Snowflake /v1/cloud 404 · exec-score saturation · SPDX 3.0 non-canonical · in-memory sort ceiling
Validated vs main @ eb58d36bb (audit P2/scale, grouped, lower priority). (1) Snowflake excluded from _INVENTORY_PROVIDERS/_CIS_PROVIDERS → 404 on /v1/cloud/{provider} inventory+CIS (cloud.py:73-74,183-185) despite the per-vendor drill north-star. (2) Exec penalty min(100,…) floors to 0/F by ~9 criticals — no discrimination at scale (exec_score.py:95,330-331). (3) SPDX 3.0 context non-canonical (.jsonl not .jsonld, 3.0.0 vs SPDX-3.0; spdx_fmt.py:16,242-255). (4) In-memory backend sorts whole tenant per read (compliance_hub_store.py:697-701) — demo backend only; document/guard the ceiling.

andres-linero:
## Summary
Two safe, self-contained items from the audit backlog #3967:

- **Snowflake CIS benchmark over REST** — `GET /v1/cloud/snowflake/cis-benchmark` returned **404** even though `snowflake_cis_benchmark.run_benchmark` is a real backend and the CLI/MCP already expose `cis-benchmark --provider snowflake`. Added `snowflake` to the REST CIS provider set with its own dispatch branch (env-resolved account/user, no region/profile scoping) and provider label. When the connector/credentials are absent, `run_benchmark` raises `CloudDiscoveryError` and the route degrades to the same **HTTP-200 `unavailable`** envelope AWS/Azure/GCP use — never 404, never 500. Brings REST to parity with CLI/MCP.
- **Guard the in-memory compliance-hub sort ceiling** — `InMemoryComplianceHubStore` copies and re-sorts a tenant's entire row list on every paged read (O(n log n) per request, independent of page size). Fine for the demo/single-node backend it's scoped to; it degrades on a large tenant. Above a documented `IN_MEMORY_SORT_CEILING` it now logs a **one-time-per-tenant** warning steering operators to the SQLite/Postgres backend (which sort in the query plan) instead of degrading silently. No change to results.

## Related Issues
Addresses #3967 (Snowflake CIS 404 + in-memory sort ceiling). The two remaining items in that grouped backlog are **not** in this PR and are tracked in a comment on the issue: the Snowflake **inventory** half needs a net-new discovery backend, and the exec-score saturation item is a scoring-model change that overlaps the in-flight exec/overview work in #3974 — so this PR does not auto-close #3967.

## Test plan
- [x] `ruff check` clean; `ruff format` clean on all changed lines (only unrelated pre-existing lines differ under local ruff 0.15.8 vs CI 0.15.21)
- [x] `mypy` clean on changed source files
- [x] No API route/Pydantic-model change: `{provider}` stays a plain path string with no new query params, so the generated OpenAPI spec is byte-identical — no regeneration needed
- [ ] `pytest tests/ -x -q` — **could not run in this sandbox** (PyPI egress is blocked 403 and the app's runtime deps, e.g. `rich`/`pydantic`, are not in the local cache, so the venv can't be built). Tests were written against the existing `test_api_cloud_surface_parity` and `test_compliance_hub_store_pagination` patterns and reviewed by hand.

### Tests added
- `test_api_cloud_surface_parity.py`: `/v1/cloud/snowflake/cis-benchmark` is wired (200 + `unavailable` envelope when the connector is absent), not 404; still enforces the read role (403 for `viewer`).
- `test_compliance_hub_store_pagination.py`: sort-ceiling guard warns once per tenant above the ceiling and stays silent below it, with paged results unchanged.

## Breaking changes
- [x] None. Snowflake CIS moves from 404 → a real/unavailable response; the sort-ceiling change is a log-only guard with no result change.

## Checklist
- [x] No secrets, credentials, or PII in diff (Snowflake creds are env/reference-only, never stored or logged)
- [x] Backend, API, docs (CHANGELOG), and tests aligned
- [x] Security review: read-only benchmark, role-gated route unchanged, degrade-to-`unavailable` avoids exception/stack exposure

## Exceptions
- None

---
_Generated by [Claude Code](https://claude.ai/code/session_013i7mYYrrizpz1DVJCiNGu3)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Versions follow [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+- **Snowflake CIS benchmark reachable over REST** (#3967): `GET /v1/cloud/snowflake/cis-benchmark` no longer 404s — it runs the real `snowflake_cis_benchmark` and, when the connector/credentials are absent, degrades to the shared HTTP-200 `unavailable` envelope like AWS/Azure/GCP, bringing the REST surface to parity with the CLI/MCP `cis-benchmark --provider snowflake`.
+- **In-memory compliance-hub sort ceiling guarded** (#3967): the process-local demo backend re-sorts a tenant's whole row list per paged read; past a documented ceiling it now warns once per tenant (steering operators to the SQLite/Postgres backend) instead of degrading silently. No behavior change to results.
+
 ## [0.95.0] - 2026-07-13
 
 Enterprise auth, cloud-connect, and interop release: browser OIDC SSO, actionable cloud connections, compliance honesty, refreshed dashboard/branding, and a large security-hardening batch on top of 0.94.2.
--- a/src/agent_bom/api/compliance_hub_store.py
+++ b/src/agent_bom/api/compliance_hub_store.py
@@ -21,6 +21,7 @@ with ``ingested_at`` so future expiry / TTL work has a key.
 
 from __future__ import annotations
 
+import logging
 import sqlite3
 import threading
 from collections.abc import Callable, Iterable, Sequence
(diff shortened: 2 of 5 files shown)


msaad00:
Superseded by #3981. The Snowflake CIS and in-memory sort-ceiling changes are preserved as a separate commit in the combined PR.

PR #3982, opened 2026-07-14
fix(compliance-hub): guard Postgres PK migration so cold start no-ops at scale

### LINKED ISSUE #3980: avail: Postgres compliance-hub cold-start 500 at scale — unconditional full-table DELETE before PK guard
Validated on current main (postgres_compliance_hub.py, `_migrate_primary_key`).

`_migrate_primary_key` runs a full-table self-join `DELETE FROM compliance_hub_findings a USING … WHERE a.tenant_id=b.tenant_id AND a.finding_id=b.finding_id AND a.ordinal > b.ordinal` **unconditionally on every store init**, BEFORE the PK-shape guard (`DO $$ … IF NOT EXISTS (pg_constraint …) ADD CONSTRAINT`) that was meant to make it a no-op. The docstring claims it's guarded, but the guard runs *after* the delete, so the delete always executes.

Impact: at ~10M rows the self-join DELETE is ~30s, exceeding the default 15s `statement_timeout` → the store never initializes → **every request 500s** (full-API outage for any large self-hosted Postgres deployment). Confirmed structurally on current main; not tracked by any issue/PR.

Fix: move the DELETE **behind** the PK-shape check (only run the dedup migration when the PK is actually the old shape / needs migrating), or gate it on a cheap 'is migration needed?' probe. Add a regression test that the migration is a true no-op when the PK is already `(tenant_id, finding_id)`.

msaad00:
## Problem

`PostgresComplianceHubStore._migrate_primary_key(conn)` ran a full-table dedup **self-join `DELETE`** unconditionally on **every** store init:

```sql
DELETE FROM compliance_hub_findings a
USING compliance_hub_findings b
WHERE a.tenant_id = b.tenant_id AND a.finding_id = b.finding_id AND a.ordinal > b.ordinal
```

The PK-shape guard (the `DO $$ … IF NOT EXISTS(pg_constraint …) ADD CONSTRAINT`) ran **after** the delete, so the delete always executed even though the docstring claimed the migration was a guarded no-op. On a large, already-migrated table (~10M rows) the O(n^2)-class self-join exceeded the default 15s `statement_timeout`, so the store never initialised and every compliance-hub request 500'd (cold-start failure).

## Fix

Make the dedup conditional by probing the current primary-key shape **first**:

- Query `pg_constraint` for the table's primary-key columns.
- If the PK is already the collapsed `(tenant_id, finding_id)` shape → **return early: no `DELETE`, no DDL.** This is the hot path for every healthy/fresh table, so init pays only a cheap catalog lookup and can never blow the statement timeout.
- Only an **old-shape** (`tenant_id, finding_id, ordinal`) or **missing** primary key triggers the dedup `DELETE` + constraint swap, preserving the original migration for genuinely-unmigrated deployments. Safe on a fresh (empty) table, whose `CREATE TABLE` already declares the collapsed PK.

## Test (TDD, no live PG required)

`tests/test_postgres_compliance_hub_migration.py` uses a connection spy (same pattern as `test_postgres_rate_limit_store.py`) that records executed SQL and answers the pkey-shape probe:

- **`test_migration_is_true_no_op_when_pk_already_collapsed`** — asserts the migration issues **only the probe** (no `DELETE`, no `ADD CONSTRAINT`) when the PK is already `(tenant_id, finding_id)`. This fails against the pre-fix code, which always issued the `DELETE`.
- **`test_migration_dedups_and_swaps_pk_on_old_shape`** — legacy `(tenant_id, finding_id, ordinal)` key still gets deduped + the collapsed PK added.
- **`test_migration_runs_when_primary_key_missing`** — no PK at all still dedups + adds the key (safe fallback).

## Verification

- `pytest tests/test_postgres_compliance_hub_migration.py` → 3 passed
- `pytest tests/test_compliance_hub.py test_compliance_hub_store_pagination.py test_compliance_hub_ingest.py test_postgres_store.py` → 164 passed
- `ruff` + `mypy` clean on the touched module
- `make preflight` green
- No live Postgres available in this environment; robustness rests on the SQL-order connection-spy unit test.

Scope: only `src/agent_bom/api/postgres_compliance_hub.py` and its new test.

Closes #3980

### DIFF
--- a/src/agent_bom/api/postgres_compliance_hub.py
+++ b/src/agent_bom/api/postgres_compliance_hub.py
@@ -356,14 +356,37 @@ class PostgresComplianceHubStore:
 
     @staticmethod
     def _migrate_primary_key(conn: Any) -> None:
-        """Collapse the primary key to ``(tenant_id, finding_id)`` (idempotent).
+        """Collapse the primary key to ``(tenant_id, finding_id)`` (true no-op once done).
 
         Pre-idempotency deployments keyed on ``(tenant_id, finding_id, ordinal)``
         so every resend of the same finding appended a fresh row. Dedup existing
-        duplicates (keep the lowest ordinal), then swap the primary key. Guarded
-        so it is a no-op once the collapsed key is in place.
+        duplicates (keep the lowest ordinal), then swap the primary key.
+
+        The dedup DELETE is a full-table self-join — O(n^2)-class on a large
+        table — so we must never run it on an already-migrated store. Probe
+        ``pg_constraint`` first and return early when the collapsed key is
+        already in place: no DELETE, no DDL. Only a genuinely old-shape (or
+        missing) primary key triggers the dedup + constraint swap. Without this
+        guard the self-join ran on every store init and blew the default 15s
+        ``statement_timeout`` at scale, so init 500'd on every request (#3980).
         """
-        # Drop duplicates ahead of the unique key, keeping the original ingest.
+        pk_row = conn.execute(
+            """
+            SELECT string_agg(a.attname, ',' ORDER BY array_position(c.conkey, a.attnum))
+              FROM pg_constraint c
+              JOIN pg_attribute a ON a.attrelid = c.conrelid AND a.attnum = ANY(c.conkey)
+             WHERE c.conrelid = 'compliance_hub_findings'::regclass
+               AND c.contype = 'p'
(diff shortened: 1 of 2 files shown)


msaad00:
Superseded by #3987. This PR's change is preserved as a separate signed commit in the combined replacement.

Issue #3931, opened 2026-07-13
Epic: dashboard UX pass — density, IA, and live scan status


msaad00:
From a live hands-on session driving the real platform (connected a real AWS account, ran a scan end-to-end). The flows work; the presentation is dense-band-heavy and under-communicates state. Bring data-heavy pages to a clean, scannable, interactive bar.

## Cross-cutting pattern
Pages are built as **stacked full-width horizontal bands**, which crowds and buries the actual content. Target pattern: **compact top toolbar + left filter rail + one dominant content canvas**, with **progressive disclosure** (collapse operator/advanced controls by default), **column chooser / add-remove**, **collapse-expand groups**, and **never a default empty state when data exists**. Content is the hero; controls recede until needed.

## Workstreams
- Security/Lineage graph IA + interactivity (linked issue)
- Scan-jobs live status + animated pipeline DAG (linked issue)
- Findings table density + column chooser (linked issue)
- (Related) Scalable cloud onboarding (separate epic)

msaad00:
**Global rule (reinforced):** on ANY data page, filters/scope/view-controls must NOT stack as full-width horizontal bands that push the graph/table/content below the fold. Collapse them into a compact top toolbar + a left filter rail (or a single collapsible controls row); the content the user came to see (graph, findings, DAG) is above the fold by default. Also raise **dark-mode contrast** — several surfaces (graph/DAG especially) read as too dark/low-contrast/flat.

PR #3983, opened 2026-07-14
fix(ui): auto-fit + node-size cap on selected exposure-path graph

msaad00:
## Problem

The Security-Graph **selected exposure-path** SVG framed its `viewBox` to a
tiny content box and then stretched it to fill the canvas, so a trivial
2-node / 1-hop path rendered as **two enormous boxes filling the whole
canvas**. Larger paths could also overflow. There was no fit-to-content
scaling and node boxes were a fixed oversized dimension.

## Fix

Extracted the graph layout into a small, testable helper
(`ui/lib/exposure-path-graph-layout.ts`) and added:

- **Auto-fit**: the `viewBox` is the tight bounding box of every node
  position plus a uniform margin, so any path frames its own content — a
  1-hop path is compact, a many-hop path scales down to fit, nothing
  overflows.
- **Render-size cap**: the SVG is capped at its natural (1x) pixel width and
  centred (`max-width` + `mx-auto` + `preserveAspectRatio="xMidYMid meet"`),
  so a small graph renders compact instead of being stretched up into giant
  boxes; larger boards still shrink-to-fit via `width: 100%`.
- **Node-size scaling**: box size interpolates from the full design size
  (<= 3 nodes) down to a readable floor (>= 9 nodes), with the label wrap
  width scaled to match so the key entity is never truncated.

Preserves the drill-to-drawer behaviour, works in light + dark, uses design
tokens only. Shared node role-style palette extracted to
`exposure-path-graph-style.ts`.

## Tests

New `tests/exposure-path-graph-layout.test.ts` covers the fit math:
few-node paths stay at full size and never larger; node size shrinks
monotonically to a floor as the path grows; a 2-node path can't fill the
canvas; every node is bounded inside the `viewBox` (no overflow) for
1/2/3/6/10/20-node paths; a 10-node path wraps into rows.

## Verify

From `ui/`: `npm ci` · `npm run typecheck` · `npm run build` · `vitest`
(exposure-path layout + command-center + neighbor-explorer) all green;
eslint clean on touched files.

## Notes

The scan-pipeline graph rendering as a flat line rather than a real
branching DAG (feedback item V) is a separate, larger change and is **not**
in this PR.

Refs #3931 (graph-UX epic) · feedback items M/W (auto-fit + node sizing).

### DIFF
--- a/ui/components/exposure-path-command-center.tsx
+++ b/ui/components/exposure-path-command-center.tsx
@@ -15,6 +15,8 @@ import {
   Wrench,
 } from "lucide-react";
 import { pathDisplayTitle, pathFixLabel, type ExposureEntityRole, type ExposurePath } from "@/lib/exposure-path";
+import { GRAPH_ROLE_STYLE } from "@/lib/exposure-path-graph-style";
+import { buildPathGraphLayout, wrapGraphText, truncateGraphText } from "@/lib/exposure-path-graph-layout";
 import { ExposurePathNeighborExplorer } from "@/components/exposure-path-neighbor-explorer";
 
 export interface ExposurePathCommandAction {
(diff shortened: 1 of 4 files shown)


msaad00:
Superseded by #3987. This PR's change is preserved as a separate signed commit in the combined replacement.

PR #4028, opened 2026-07-15
fix: pre-release self-host fixes — quickstart works out of the box

msaad00:
The 0.95.0 self-host quickstart did not come up out of the box. Bug-hunt reproduced two P0s on the published images, plus P1/P2 friction. Fixes below, each with the reproduced-break → fixed reasoning.

### P0-1 · Dockerfile: create + chown the state dir (unbricks the API)
`Dockerfile` did `USER abom` (uid 100) but never created/chowned `/home/abom/.agent-bom`. The pilot named volume `agent-bom-pilot-data` (and platform `agent-bom-state`) mounts over that path root-owned → non-root `abom` cannot create `jobs.db`/`vulns.db` → API crash-loops → UI never starts (`depends_on: service_healthy`).
Fix: before `USER abom`, `RUN mkdir -p /home/abom/.agent-bom && chown -R abom:abom /home/abom`. A fresh named volume inherits the (now abom-owned) image mount-point, so the app can write.

### P0-2 · Compose: point the UI at the API service DNS
`ui.NEXT_PUBLIC_API_URL` was `http://localhost:8422` in pilot/fullstack/platform. The UI's Node server (per `ui/next.config.ts`) proxies `/v1,/health,/ws` to that URL from *inside* the compose network — `localhost` there is the UI container itself, so every API call failed. Set it to the API service DNS `http://api:8422` (the `api` service is the canonical compose network alias in all three files).

### P1-1 · Lead the quickstart with secret generation
platform/fullstack need six mounted secret files (only `*.example` committed); `docker compose up` hard-fails with a bare "secret file not found" if skipped. Made secret-gen step 1 in both compose headers and the docs (`docs/DEPLOY_PLATFORM.md`, `site-docs/deployment/quickstart.md`), and added `make secrets` / `make platform-up` / `make fullstack-up` (secrets target runs `hosted_poc_preflight.py --write-secret --skip-compose`).

### P1-2 · Drop the dead fullstack build-arg
`docker-compose.fullstack.yml` passed `NEXT_PUBLIC_API_URL` as `build.args` but `ui/Dockerfile` has no matching `ARG` → silently ignored. Removed it; the P0-2 runtime env is what matters.

### P2 · Remove obsolete `version: '3.8'`
Dropped from fullstack + platform (modern Docker warns on it).

## Verification (built + booted — real repro, then torn down)
Built the API + UI images with the fixes and brought up the pilot stack:
- **P0-1:** `agent-bom-pilot-api` reached `Up (healthy)` on a fresh named volume; `jobs.db` created and owned by `abom` (uid 100) — no crash loop. Pre-fix, the same mount is root-owned and the write fails.
- **P0-2:** from inside the compose network, `http://api:8422/health` returns **HTTP 200** — the exact origin the UI's Node server proxies to. With the old `localhost:8422` the UI server would hit itself (connection refused).
- **Static:** `docker compose -f deploy/docker-compose.{pilot,fullstack,platform}.yml config` all exit 0 with no `version` warning; rendered `ui` env is `http://api:8422` in all three; Dockerfile chown line is immediately before `USER abom`.

## Security defaults confirmed untouched
`AGENT_BOM_NO_AUTH_ROLE=analyst`, file-mounted Docker secrets, non-superuser `agent_bom_app` DB role, loopback (`127.0.0.1`) binds, internal-only Postgres network (`internal: true`), connections_key auto-seed — all intact.

## ⚠️ Overlap note for reconciliation
The in-flight `control-plane-secrets-files` branch also edits `deploy/docker-compose.{fullstack,platform,product}.yml` and `deploy/secrets/`. This PR's compose changes were kept minimal and surgical (only the `ui` `NEXT_PUBLIC_API_URL` line, the dead build-arg removal, the `version:` line, and the header-comment quickstart steps). Please reconcile at merge — no `deploy/secrets/` or Postgres-role changes were made here, and `docker-compose.product.yml` was intentionally left untouched to reduce overlap.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,6 +68,12 @@ RUN addgroup -S abom && adduser -S -G abom abom
 WORKDIR /workspace
 RUN chown abom:abom /workspace
 
+# Pre-create the state dir and hand ownership to abom (uid/gid of the abom
+# user) so a named volume or bind mount at /home/abom/.agent-bom is writable.
+# Without this the mount lands root-owned and abom (non-root) cannot create
+# jobs.db / vulns.db, crash-looping the API.
+RUN mkdir -p /home/abom/.agent-bom && chown -R abom:abom /home/abom
+
 USER abom
 
 ENV PYTHONUNBUFFERED=1
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help install test lint preflight preflight-fix docker-build docker-run scan clean build-ui analytics dev check-dupes clean-dupes
+.PHONY: help install test lint preflight preflight-fix docker-build docker-run scan clean build-ui analytics dev check-dupes clean-dupes secrets platform-up fullstack-up
 
 help:  ## Show this help message
 	@echo 'Usage: make [target]'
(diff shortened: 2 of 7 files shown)


msaad00:
Temporarily closing and reopening to retrigger required checks for current head 8f26591; this SHA did not receive a check suite.

agent-bom (approved):

andres-linero (approved):

Issue #4095, opened 2026-07-17
Epic: post-merge accuracy, deployment evidence, and operator workflow hardening


msaad00:
## Objective

Close the three remaining workstreams from the original post-merge accuracy audit. This issue is a bounded historical umbrella, not the tracker for every later CNAPP or runtime program.

## Original workstreams — current status

- [x] #4092 — posture, enrichment, graph, compliance, and remediation-output accuracy; closed by #4096
- [x] #4093 — release proof, air-gap behavior, and deployment health; closed by #4098
- [x] #4094 — prioritized risk campaigns and unified investigation workflow; closed by #4100
- [x] #4101 — bounded ATT&CK/ATLAS evidence mapping; closed by #4102
- [x] #4103 — PostgreSQL store parity; closed by #4110/#4140
- [x] #4104 — CycloneDX reference uniqueness; closed by #4109
- [x] #4105 — atomic warehouse export publication and credential validation; closed by #4122
- [x] #4106 — non-blocking compliance-hub reads and bounded continuation; closed by #4133/#4146
- [x] #4107 — Snowflake freshness and unevaluable-control semantics; closed by #4124
- [ ] #4108 — versioned ATT&CK/ATLAS catalogs and typed multi-hop attack-path mappings
- [ ] #4120 — benchmark-catalog provenance and deeper provider coverage
- [ ] #4075 — production streaming graph producer and bounded write-path memory

Progress: **9/12 original workstreams complete**.

## Successor epics — tracked separately

The following programs were discovered or expanded after this audit. They do not silently enlarge this issue's completion contract:

- #4131 — production CIEM evidence consumption
- #4134 — evidence-complete Kubernetes posture
- #4151 — verifiable MCP server attestations
- #4152 — durable profile-aware runtime gateway activity
- #4156 — large-estate attack-path campaigns
- #4157 — Azure Blob and production database DSPM
- #4158 — cross-cloud disk scanning and runtime workload evidence

## Delivery contract

- Use regression-first TDD and reproduce every defect before implementation.
- Preserve read-only, agentless behavior and connect-once credentials.
- Validate artifacts, tenant isolation, invalid/no-data states, large-estate behavior, deployments, and product surfaces in proportion to each workstream.
- Keep independently scoped security and performance changes in their owning series; do not combine unrelated blast radii merely to close this umbrella.

Close #4095 only after #4108, #4120, and #4075 are complete and the merged result is re-audited.

msaad00:
[...]

PR #4140, opened 2026-07-17
fix(postgres): restore bootstrap, snapshot, and edge parity

### LINKED ISSUE #4136: fix(graph): restore Postgres snapshot JSON and edge-activity parity
## Problem

Fresh #4132 validation found two production Postgres parity defects:

1. `deploy/supabase/postgres/init.sql` creates `graph_snapshots.analysis_status` as JSONB, while `PostgresGraphStore._init_tables` declares/adds it as TEXT. Psycopg3 returns JSONB as a dict, but `load_graph` calls `json.loads` unconditionally, so any non-empty persisted analysis status can crash graph reads on Supabase/init.sql bootstraps. `risk_summary` uses the same bare decode pattern and must be audited/fixed together.
2. Retired-edge persistence diverges: PostgreSQL changes `activity_id` from Create(1) to Close(3), while SQLite preserves Create(1). Because activity ID participates in edge-change/OCSF output, identical scans produce backend-dependent evidence.

## Acceptance

- [ ] reproduce non-empty JSONB row decoding with dict-returning psycopg behavior
- [ ] accept both existing TEXT JSON and native JSONB objects without raw errors
- [ ] align canonical bootstrap/runtime column types and migration behavior without breaking existing deployments
- [ ] lock `risk_summary` and `analysis_status` parity
- [ ] choose and document one retired-edge `activity_id` invariant across SQLite/Postgres
- [ ] assert `/v1/graph/edges/changes` and OCSF fingerprints are backend-consistent
- [ ] run SQLite and live Postgres graph/migration/API regressions

Regression found during post-merge validation of #4110 and #4132.

Addresses #4095.

### LINKED ISSUE #4138: fix(deploy): make Alembic Postgres baseline bootstrap psycopg-safe
## Problem

Live PostgreSQL 16 verification for #4136 exposed a pre-existing baseline-bootstrap failure. `20260416_01_control_plane_baseline.py` passes the full `init.sql` string to SQLAlchemy `exec_driver_sql`. Under psycopg3, PL/pgSQL `format(... %L ...)` tokens in the role-password block are interpreted as client parameter placeholders, causing `ProgrammingError: only %s/%b/%t allowed, got %L` before a fresh Alembic-only database can bootstrap.

Docker init may remain unaffected because psql executes the file directly; the Alembic deployment path is not equivalent.

## Acceptance

- [ ] reproduce against the shipped PostgreSQL 16 image and psycopg3
- [ ] execute baseline SQL without client placeholder interpolation while preserving PL/pgSQL `%L` semantics
- [ ] preserve app-role password handling and avoid logging/interpolating secrets in Python
- [ ] prove empty-database Alembic upgrade to head
- [ ] prove re-run/idempotency and existing-database forward upgrade
- [ ] retain Docker/psql bootstrap parity
- [ ] add a CI contract that exercises the actual baseline migration rather than stamping past it

Found while validating #4136; keep separate from graph snapshot parity.

msaad00:
## Summary

- make the Alembic baseline and partition migration driver-safe so both supported Postgres bootstrap paths reach the current head
- decode graph snapshot JSON from psycopg-native mappings or serialized text, align snapshot storage, and preserve a reversible migration path
- standardize retired-edge evidence on OCSF Close(3) across SQLite and Postgres, with API and backend parity regressions

## Verification

- combined migration, schema, streamed-persistence, SQLite parity, and graph API suite: 152 passed
- owning graph/Postgres/OCSF suite before consolidation: 755 passed
- live PostgreSQL 16 graph migration: native JSONB read, 20260717_02 to 03 conversion, save/load, retired-edge Close(3), downgrade and replay
- shipped postgres:16-alpine bootstrap: empty Alembic database to head plus rerun; init.sql baseline stamp plus forward upgrade
- app-role Postgres storage contract: 5 passed, 1 expected privilege skip
- Ruff check and format, mypy, exception-sanitization guard, git diff check
- make preflight: OpenAPI, v1 schemas, product surface, release consistency, env references, and SDK patterns clean

## Notes

- readers remain tolerant of native JSONB mappings during mixed-version rollout; snapshot writes use the existing serialized graph-store contract
- the snapshot type conversion takes a table lock and should run in a maintenance window on hot installations
- both bootstrap paths clear the transient password GUC and leave bootstrap owners without SUPERUSER or BYPASSRLS
- #4141 and #4142 were closed after their unique coverage and bootstrap commit were consolidated here

Closes #4136
Closes #4138
Addresses #4095

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,7 +98,7 @@ jobs:
           if printf '%s\n' "$changed" | grep -Eq '^(site-docs/|mkdocs\.yml$|\.github/workflows/(docs|ci)\.yml$|\.github/actions/setup-python/)'; then
             docs=true
           fi
-          if printf '%s\n' "$changed" | grep -Eq '^(src/agent_bom/api/postgres_|src/agent_bom/api/storage_schema\.py|tests/test_postgres_)'; then
+          if printf '%s\n' "$changed" | grep -Eq '^(src/agent_bom/api/postgres_|src/agent_bom/api/storage_schema\.py|tests/test_postgres_|deploy/supabase/postgres/)'; then
             postgres=true
           fi
           if printf '%s\n' "$changed" | grep -Eq '^(deploy/endpoints/|scripts/(build-[^/]+|render_homebrew_formula\.py)|src/agent_bom/cli/_runtime\.py|tests/test_cli_runtime\.py|pyproject\.toml|uv\.lock|\.github/actions/setup-python/|site-docs/deployment/aws-company-rollout\.md$)'; then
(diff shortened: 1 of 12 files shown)


msaad00:
Consolidation note: #4141 is folded here rather than merged independently. This branch remains the single graph P1 fix, aligns the snapshot columns with the existing graph-store serialized-TEXT contract, keeps mixed TEXT/JSONB readers for rollout compatibility, and preserves live migration, rollback, API, and OCSF parity verification.

andres-linero (approved):

agent-bom (approved):

PR #4141, opened 2026-07-17
fix(graph): guard JSONB analysis_status reads and reconcile retired-edge activity_id across backends

msaad00:
Two contained graph-backend correctness bugs, one PR. TDD: both regressions have a test that fails RED before the fix and passes after.

## Fix 1 — P1: `analysis_status`/`risk_summary` crash graph reads on the JSONB schema

**Root cause.** `deploy/supabase/postgres/init.sql` and the enterprise migration `20260717_02_graph_analysis_status.py` declare `graph_snapshots.analysis_status` (and `risk_summary`) as **JSONB**, but the app's own `_init_tables` created legacy tables as **TEXT** — a schema-type divergence. The Postgres reads did bare `json.loads(...)`:
- `postgres_graph.py` `load_graph` (`analysis_status`)
- `list_snapshots` (`risk_summary` + `analysis_status`)
- snapshot-detail (`analysis_status`)

Under psycopg3 a JSONB column returns an **already-parsed `dict`**, so `json.loads(dict)` raises `TypeError` and the graph read crashes — and only when the status is non-empty (an empty `{}` is falsy so it fell back to `"{}"`), i.e. exactly when attack-path fusion recorded a status.

**Fix.** Added a `_json_column_to_dict` guard that is robust to BOTH a `str` (TEXT column) and a `dict` (JSONB column) — mirroring the existing `isinstance(x, dict)` guard in `postgres_audit.py`. Applied at all three `analysis_status` read sites and the sibling `risk_summary` read.

**Schema-alignment choice.** The canonical schema (`init.sql` + the existing `20260717_02` migration) is already **JSONB**; the app's `_init_tables` was the lone TEXT outlier. I aligned `_init_tables` (both the `CREATE TABLE` and the `ADD COLUMN` ALTER) to JSONB to match, so new app-path tables are consistent with the canonical schema. No new migration is needed (the JSONB migration already exists). Existing legacy TEXT tables are untouched (`CREATE/ADD COLUMN IF NOT EXISTS` no-op on an existing column) and remain correct because the `isinstance` guard handles both types. The write path already stores `json.dumps(...)` strings, which PostgreSQL assignment-casts into JSONB (unchanged, already how JSONB production DBs worked).

**How verified.** New tests simulate psycopg3's dict return for JSONB and drive the real `load_graph`/`list_snapshots` — RED before (`TypeError: the JSON object must be str, bytes or bytearray, not dict`), green after.

## Fix 2 — P1-class: retired-edge `activity_id` parity (SQLite ≠ Postgres)

**Root cause.** On retiring an edge dropped from a new snapshot, Postgres stamps `activity_id = CASE WHEN activity_id = 1 THEN 3 ELSE activity_id END` (OCSF Create→Close), but SQLite only set `valid_to` and left Create(1). `activity_id` is in the edge-change fingerprint on both backends and flows into `edges_removed` → the change feed / OCSF `type_uid`, so the same scan sequence yielded a different diff by backend.

**Fix.** Aligned the SQLite retire UPDATE to also stamp Create(1)→Close(3), matching Postgres (the more correct OCSF change-feed semantics — a retired edge is Closed).

**How verified.** New cross-backend parity test: retire the same edge on SQLite end-to-end and assert `edges_removed[0]["activity_id"] == 3` (RED: was `1`), plus a structural assertion that both stores issue the identical `CASE` stamp (RED before on the SQLite side).

## Scope / safety
Backend-only (graph stores). No API/UI contract shape change: reads return the same `analysis_status` dict; `activity_id` values only reconcile. `tenant_id`/RLS scoping preserved on every touched query; no new UNIQUE keys.

## VERIFY (all reproduced, green)
- New tests: 4 RED before → 4 passed after.
- Touched suites (both backends): `test_graph_api test_graph_delta_digest test_graph_backend test_postgres_graph_streaming test_postgres_graph_retention test_attack_path_fusion test_graph_attack_path_fusion test_postgres_migrations test_ocsf test_graph_cloud_edge_lifecycle test_graph_roundtrip` → 182 passed.
- `ruff check` + `ruff format --check` clean; `mypy` changed files clean; `bandit -r src/ --severity-level medium` exit 0.
- `scripts/check-counts.py`, `export_openapi.py --check`, `generate_v1_schemas.py --check` all pass.

### DIFF
--- a/src/agent_bom/api/postgres_graph.py
+++ b/src/agent_bom/api/postgres_graph.py
@@ -32,6 +32,23 @@ from .postgres_common import _apply_tenant_session, _ensure_tenant_rls, _get_poo
 logger = logging.getLogger(__name__)
 
 
+def _json_column_to_dict(value: Any) -> dict[str, Any]:
+    """Normalize a JSON/JSONB column value to a ``dict``.
+
+    ``init.sql`` and the enterprise migration declare ``risk_summary`` /
+    ``analysis_status`` as ``JSONB`` while the app's own ``_init_tables`` created
+    legacy tables as ``TEXT``. Under psycopg3 a ``JSONB`` column returns an
+    already-parsed ``dict`` and a ``TEXT`` column returns a ``str``, so the read
+    must handle both to stay correct across the two schema variants. Mirrors the
+    guard already used in ``postgres_audit`` (``isinstance(x, dict)``).
+    """
+    if isinstance(value, dict):
+        return value
+    if isinstance(value, str):
+        return json.loads(value or "{}")
+    return {}
+
+
 def select_expired_snapshot_ids(
     rows: Iterable[Sequence[Any]],
     *,
@@ -304,8 +321,8 @@ class PostgresGraphStore:
                     created_at TEXT NOT NULL,
                     node_count INTEGER DEFAULT 0,
                     edge_count INTEGER DEFAULT 0,
-                    risk_summary TEXT DEFAULT '{}',
-                    analysis_status TEXT DEFAULT '{}',
+                    risk_summary JSONB DEFAULT '{}'::jsonb,
+                    analysis_status JSONB NOT NULL DEFAULT '{}'::jsonb,
                     PRIMARY KEY (scan_id, tenant_id)
                 )
                 """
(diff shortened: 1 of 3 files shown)


msaad00:
Consolidating this duplicate graph-backend fix into #4140. #4140 already carries the live PostgreSQL migration and rollback proof, API/OCSF parity coverage, and the exact #4136 closure. The combined review retains the graph store serialized-TEXT contract while keeping readers tolerant of psycopg-native JSONB mappings during rollout. Closing this single PR avoids two competing schema directions and duplicate CI.

PR #4142, opened 2026-07-17
fix(deploy): make Alembic Postgres bootstrap driver-safe

### LINKED ISSUE #4138: fix(deploy): make Alembic Postgres baseline bootstrap psycopg-safe

msaad00:
## Summary

- preserve server-side PL/pgSQL percent-format tokens through the psycopg3 baseline migration without client parameter parsing
- unwrap the SQLAlchemy connection for the shared psycopg partition helper while retaining the Alembic transaction
- exercise both supported Postgres bootstrap paths in CI: empty Alembic-to-head and init.sql baseline-to-head, including rerun, credential cleanup, and bootstrap-privilege assertions

## Verification

- shipped postgres:16-alpine: empty Alembic database to head plus idempotent rerun
- shipped postgres:16-alpine: init.sql, baseline stamp, and forward upgrade to head
- app-role Postgres storage contract: 5 passed, 1 expected privilege skip
- 120 owning and security tests passed
- Ruff check and format, mypy, actionlint, workflow timeout/release lint, exception-sanitization guard, and git diff check

## Notes

- root cause 1: SQLAlchemy supplied an empty parameter mapping, so psycopg3 interpreted PL/pgSQL percent-L tokens as client placeholders
- root cause 2: a later migration passed a SQLAlchemy Connection to a helper that intentionally implements the psycopg execute contract
- both migration paths clear the transient password GUC and leave bootstrap owners without SUPERUSER or BYPASSRLS

Closes #4138
Addresses #4095

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,7 +98,7 @@ jobs:
           if printf '%s\n' "$changed" | grep -Eq '^(site-docs/|mkdocs\.yml$|\.github/workflows/(docs|ci)\.yml$|\.github/actions/setup-python/)'; then
             docs=true
           fi
-          if printf '%s\n' "$changed" | grep -Eq '^(src/agent_bom/api/postgres_|src/agent_bom/api/storage_schema\.py|tests/test_postgres_)'; then
+          if printf '%s\n' "$changed" | grep -Eq '^(src/agent_bom/api/postgres_|src/agent_bom/api/storage_schema\.py|tests/test_postgres_|deploy/supabase/postgres/)'; then
             postgres=true
           fi
           if printf '%s\n' "$changed" | grep -Eq '^(deploy/endpoints/|scripts/(build-[^/]+|render_homebrew_formula\.py)|src/agent_bom/cli/_runtime\.py|tests/test_cli_runtime\.py|pyproject\.toml|uv\.lock|\.github/actions/setup-python/|site-docs/deployment/aws-company-rollout\.md$)'; then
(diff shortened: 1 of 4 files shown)


msaad00:
Consolidated into #4140 because the graph revision depends on this repaired Alembic bootstrap chain for a real empty-database upgrade. The combined branch preserves both signed commits, closes #4136 and #4138, and adds a base-matrix Alembic import stub after reproducing the standalone Python failure locally. #4140 will receive one authoritative combined CI run; closing this single PR avoids duplicate migrations and duplicate CI.

Issue #4106, opened 2026-07-17
perf(compliance): offload hub reads and add bounded continuation


msaad00:
## Problem

Async compliance-hub routes perform synchronous store work on the event loop. Findings pagination is also OFFSET-based and requests `offset + limit` rows before slicing, so latency and work grow with the requested page depth.

Current main already reconciles executive counts correctly; that stale claim is excluded.

## Acceptance criteria

- [ ] offload synchronous compliance findings and posture aggregate reads from the event loop
- [ ] expose bounded cursor/keyset continuation for hub findings
- [ ] preserve tenant isolation, ordering, filters, and honest total/count semantics
- [ ] retain compatibility for existing bounded first-page callers or version the contract explicitly
- [ ] add event-loop responsiveness and deep-pagination regressions
- [ ] align OpenAPI, SDK, UI continuation behavior, and operator docs

Part of #4095.

msaad00:
Fresh origin/main runtime re-audit after #4133: this issue is not fully closed.

Confirmed remaining gaps:
- stale 400-day scan: /v1/findings critical=0 under its 90-day window, while /v1/overview and /v1/posture/counts report critical=1
- two recent rescans of the same blast-radius CVE: findings dedupes to 1, overview/posture report 2
- the dashboard independently recomputes severity from the latest 10 hydrated jobs, then links that count to the findings view, so the visible count can disagree even after an API-only repair
- /hub/findings and /hub/posture are genuinely offloaded, but /posture/counts still calls exec_severity_counts inline; an injected 0.2-second body blocked the event loop for 0.215 seconds
- hub findings still uses ordinal/offset pagination; the accepted bounded cursor/keyset continuation and SDK/UI continuation are absent
- Postgres ingest_batch_atomic invalidates the overview cache but not findings_count_cache: live API reproduced count=2 with stale approximate total=1 immediately after the second atomic ingest

Positive regression evidence: the exact former Python 3.11 xdist seed passed 46/46; focused cache/order suites passed 62/62; live Postgres severity/framework aggregates match canonical list counts outside the stale invalidation case.

Remaining work should combine server-side count semantics, Postgres invalidation, async offload, cursor pagination, and dashboard consumption under one regression matrix. Addresses #4095.

PR #4143, opened 2026-07-17
fix(compliance): reconcile executive finding counts

msaad00:
## Summary

- share one canonical default-window, parent-job exclusion, and latest-wins current-state fold across findings, overview, and posture severity counts
- prevent stale compact summaries and duplicate blast-radius rescans from creating executive counts without matching drill-down rows
- invalidate every tenant findings-total cache after a committed Postgres atomic ingest

## Verification

- focused compliance, overview, and atomic-ingest matrix: 100 passed, 4 optional Postgres skips
- current-main rerun: 45 passed, 4 optional Postgres skips
- stale 400-day scan, duplicate rescan, and atomic-ingest cache regressions pass
- Ruff check and format, mypy, exception-sanitization guard, git diff check
- make preflight: OpenAPI, v1 schemas, product surface, release consistency, env references, and SDK patterns clean

## Notes

- phase 1/P1 only: dashboard count consumption, posture-counts event-loop offload, and hub cursor/keyset continuation remain open in #4106
- historical compact summaries still populate scan/domain coverage and the score floor, but no longer fabricate executive severity without a current drill-down row
- tenant isolation and the documented approximate-total contract are preserved

Addresses #4106
Addresses #4095

### DIFF
--- a/src/agent_bom/api/findings_current.py
+++ b/src/agent_bom/api/findings_current.py
@@ -1,0 +1,81 @@
+"""Canonical current-state fold for scan-produced findings.
+
+The findings list and executive count surfaces share this module so their
+default time-window, parent-job exclusion, and cross-job replacement semantics
+cannot drift independently.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Callable, Iterable
+from datetime import datetime, timezone
+from typing import Any
+
+from agent_bom.api.models import JobStatus
+
+
+def job_in_window(job: Any, since: str | None) -> bool:
+    """Return whether a job's completion timestamp is inside ``since``."""
+    if since is None:
+        return True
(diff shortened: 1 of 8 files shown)


msaad00:
CI follow-up: Python 3.11, 3.13, and 3.14 each failed only tests/test_core.py::test_api_posture_counts_with_data because its fixed 2026-01-01 timestamp is outside the new canonical 90-day window. The exact test reproduced locally before the fix. The fixture now uses current UTC time, while dedicated 400-day and duplicate-rescan regressions retain the exclusion/dedup proof; those three tests pass together. Rebased onto current main and pushed once after the original run completed.

agent-bom (approved):

andres-linero (approved):

PR #4180, opened 2026-07-18
fix: audit P0/P1 — event-loop offload, SCA matching + Go reachability honesty

msaad00:
Two independently-verified, disjoint-file fixes for the highest-severity findings from the multi-dimension audit, combined into one PR (two commit streams for revert granularity). Both were green on their own branches; the full gate was re-run on the merged result (1424 passed across `tests/api/` + `tests/cloud/` + osv).

## 1. Scale P0 — heavy sync work off the event loop
Non-negotiable: no blocking work on the async loop.
- `routes/cloud.py` — inventory + CIS handlers offload their provider-scan bodies (extracted to `_build_inventory_payload` / `_run_cis_benchmark`) via `anyio.to_thread.run_sync` under `adaptive_backpressure`, with a 429 on rejection, mirroring the existing `_build_account_summary` idiom.
- `routes/governance.py` — all six Snowflake ACCESS_HISTORY/QUERY_HISTORY/Cortex handlers offload through a shared `_offload` helper (+ `try/finally` on `conn.close()` so a connection isn't leaked on error).
- `auth.py` / `middleware.py` — `KeyStore.verify` now holds its lock only to snapshot the key list, so the ~21ms scrypt derivation runs lock-free (no reserialization; `is_usable()` still checked at match time, revocation/expiry unchanged); both `APIKeyMiddleware` and `RateLimitMiddleware._resolve_tenant_scope` offload the verify. The rate-limit path was a **second, unnamed instance** of the same defect — the only verify on the anonymous/demo request stack.
- Behavioral test proves `/health`-style coroutines stay responsive during a slow scan; payloads and error semantics byte-for-byte unchanged.

## 2. SCA matching honesty (incl. a P0)
Read-only scanner must never assert a version/CVE/fix the evidence doesn't support.
- **P0 — "no fix" on a fixed CVE, then dropped under `--exclude-unfixable`**: `db/lookup.py` now prefers the version-matched range `fixed`, falling back to the advisory rollup only when it's a *valid* version (guards the git-SHA rollup that made urllib3 CVE-2023-43804 look unfixable). `scanners/osv.py` `parse_fixed_version` now selects the fix from the affected branch that **contains** the installed version (order-independent), so CVE-2023-45803 reports 1.26.18, not the 2.x jump 2.0.7.
- **Range constraints scanned as exact pins**: `parsers/python_parsers.py` emits `==` as an installed pin but a range/inequality as a floating declaration reference (`version=unknown`, `floating_reference=True`, no fabricated purl) — applied to requirements.txt, pyproject.toml, and the pip-compile parser.
- **Bare deps resolved from the scanner host**: `package_scan.py` threads the `-p` target dir (`ScanOptions.project_dir`) so pip/npm/go resolve from the target, never the scanner's env/cwd; no target install ⇒ nothing stamped `installed`.
- **Reachability over-claim**: declaration-manifest packages set `reachability_evidence="declaration_only"`, firing the existing `models.py` guard so reachability reads `unknown`, not `likely`.
- **CVSS score/vector mismatch**: `db/sync.py` adopts the vector only alongside the score from the same severity entry.

## How verified
- New TDD suites reproduce each defect before the fix: `tests/test_vuln_matching_honesty.py` (13), `tests/api/test_cloud_scan_offload.py`, `test_governance_offload.py`, `test_api_key_verify_offload.py`.
- Merged-result gate: `uv run pytest` over `tests/api/` + `tests/cloud/` + osv accuracy = **1424 passed**; `ruff` clean; `mypy` clean on all 9 changed source files.
- Live CLI smoke: `check urllib3@1.26.4` now reports 1.26.17 / 1.26.18 / 1.26.19 (same-branch, was "no fix"/2.x jumps); `--exclude-unfixable` retains CVE-2023-43804; range specs emit `version=unknown` with no purl.

## Reviewer note
One assertion in the offload test was corrected (last commit): it asserted a `provider`/`status:unavailable` shape the endpoint never returns — verified identical on `origin/main` — so it now asserts the offload happened and the real payload flows back.

### DIFF
--- a/src/agent_bom/api/auth.py
+++ b/src/agent_bom/api/auth.py
@@ -471,8 +471,15 @@ class KeyStore:
         return [k for k in keys if k.tenant_id == tenant_id]
 
     def verify(self, raw_key: str) -> ApiKey | None:
+        # Hold the lock only long enough to snapshot the key list; the expensive
+        # scrypt derivation in ``verify_api_key`` then runs lock-free so it
+        # neither stalls the caller under the lock nor serializes concurrent
+        # verifies (each worker thread derives independently). ``is_usable()`` is
+        # still evaluated at match time, so revocation/expiry semantics are
+        # unchanged.
         with self._lock:
-            return verify_api_key(raw_key, self._keys)
+            keys_snapshot = list(self._keys)
+        return verify_api_key(raw_key, keys_snapshot)
 
     def has_keys(self) -> bool:
         with self._lock:
--- a/src/agent_bom/api/middleware.py
+++ b/src/agent_bom/api/middleware.py
@@ -17,6 +17,8 @@ from collections.abc import Sequence
 from functools import lru_cache
 from typing import TYPE_CHECKING, cast
 
+import anyio.to_thread
+
 from agent_bom import __version__
 from agent_bom.api.auth import get_key_store
 from agent_bom.api.browser_session import (
@@ -1252,7 +1254,10 @@ class APIKeyMiddleware(BaseHTTPMiddleware):
 
         store = get_key_store()
         if store.has_keys():
-            api_key = store.verify(raw_key)
+            # ``store.verify`` runs a ~21ms scrypt derivation (in-memory store) or
+            # a blocking DB read (Postgres store); offload it to a worker thread so
+            # it never stalls the async event loop while unrelated requests wait.
+            api_key = await anyio.to_thread.run_sync(store.verify, raw_key)
             if api_key:
                 required = self._required_role(request.method, request.url.path)
                 required_role = Role(required)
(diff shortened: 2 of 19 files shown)


andres-linero (dismissed):

agent-bom (dismissed):

msaad00:
Folded in #4181 (Go SAST: mask comments/string literals so sink/LLM tokens in `//`, `/* */`, "...", and backtick raw strings no longer inflate Go findings/reachability) — same accuracy class as the SCA matching fixes here. Also corrected the CIS-offload test to be environment-invariant: the offloaded payload shape depends on whether the cloud SDK is installed (evaluated vs honest `unavailable` shape), so it now asserts only the offload + tenant invariants, fixing the CI test failure.

msaad00:
Folded in the UI bundle-budget headroom (3536→3552 KiB) so UI Validate passes standalone — main was sitting at the exact bundle ceiling, failing UI Validate on any PR that touches `api/routes` even with no UI change. This backend diff changes no UI; the bump is measured CI-variance headroom (largest-chunk and shared-runtime budgets unchanged). #4188 closed in favor of this.

agent-bom (approved):

andres-linero (approved):

PR #4181, opened 2026-07-18
fix(sast): mask comments and string literals in Go source analysis

msaad00:
## What
`scan_go_file` was the only one of seven regex source parsers running its dangerous/LLM pattern passes and its call-site extraction against raw Go source. Sink tokens (`exec.Command`, `os.WriteFile`, `template.HTML`) and LLM-client tokens appearing inside `//` and `/* */` comments, `"..."` strings, or backtick raw strings were reported as real dangerous/LLM capability calls — even with no matching import — inflating Go findings and reachability.

## Fix
- `_go_call_sites` and the dangerous/LLM pattern pass in `scan_go_file` now run against `mask_line_comments_and_strings(..., backtick_strings=True)`, mirroring the php/ruby/java/rust/swift/csharp call-site passes. Prompt/import extraction still reads raw source (it needs literal contents).
- Added an **opt-in** `backtick_strings` flag to the shared masker for Go raw strings. It defaults off, so the other six parsers call the helper unchanged (a backtick means something different in Swift/Ruby/PHP). Masking preserves character positions and newlines, so line numbers stay accurate.

## How verified
- New tests (red → green): sink/LLM tokens in comments, `"..."` strings, and backtick raw strings produce zero findings; positive controls confirm real calls are still flagged.
- Parser regression suite: `205 passed`. New Go tests: `4 passed`. `ruff check`/`format`: clean. `mypy` on both changed source files: clean. Ran the other six parsers' suites — no regression from the masker change.

### DIFF
--- a/src/agent_bom/ast_go.py
+++ b/src/agent_bom/ast_go.py
@@ -24,6 +24,7 @@ from agent_bom.ast_signal_utils import (
     check_prompt_risks,
     classify_prompt_type,
 )
+from agent_bom.ast_source_mask import mask_line_comments_and_strings
 
 if TYPE_CHECKING:
     from collections.abc import Mapping, Sequence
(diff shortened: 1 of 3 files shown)


msaad00:
Folded into #4180 (same reachability-accuracy class; merged there, gate re-run green on the combined result).

PR #4182, opened 2026-07-18
fix(identity): fail closed on partial IAM evidence and unpinned JWT audience

msaad00:
Two independent identity/authz hardening fixes, fail-closed in the safe direction.

## 1. Admin-equivalence heuristic suppressed under partial IAM evidence
`apply_effective_permissions` treated any non-empty normalized-policy set as authoritative. Because the evaluator returns INDETERMINATE when any document is only PARTIAL (a single malformed statement), a role with `AdministratorAccess` plus one malformed policy evaluated as "not admin", and that non-authoritative verdict suppressed the name/keyword heuristic — under-flagging a real admin.

**Fix:** the heuristic is suppressed only when evidence is AUTHORITATIVE (every collected document COMPLETE). PARTIAL/incomplete evidence falls through to the heuristic (fails toward flagging). A COMPLETE, authoritatively-not-admin result still suppresses — no new false positives.

## 2. Inbound JWT not audience-pinned when jwks_uri set directly
With a direct `jwks_uri` and no `expected_audience`, `resolve_agent_id` decoded with `verify_aud=False`, so a validly-signed token minted by the same IdP keys for a different relying party was accepted (audience confusion).

**Fix:** a directly-configured `jwks_uri` now requires `expected_audience`; without it, verification is refused with a clear operator message. The `oidc_issuer` path (issuer pinned) and the explicit-audience path are unchanged. No production config sets `jwks_uri` directly.

## How verified
- TDD red→green per defect, plus positive controls (complete not-admin still suppresses; matching aud accepted, wrong aud rejected).
- Existing tests that encoded the old insecure JWT contract were updated to the correct fail-closed expectation.
- Combined suite: `152 passed`. `ruff` check + format clean; `mypy` clean on both changed source files.

### DIFF
--- a/src/agent_bom/agent_identity.py
+++ b/src/agent_bom/agent_identity.py
@@ -15,9 +15,14 @@ Two supported token formats:
   (``{token: agent_id}``).
 
 Policy keys:
-- ``jwks_uri``: URL to a JWKS endpoint for signature verification
+- ``jwks_uri``: URL to a JWKS endpoint for signature verification. When set
+  directly, ``expected_audience`` MUST also be set or JWT verification fails
+  closed — otherwise a validly-signed token minted for a different relying
+  party by the same IdP keys would be accepted (audience confusion).
 - ``oidc_issuer``: OIDC issuer base URL; ``jwks_uri`` auto-discovered via
-  ``{issuer}/.well-known/openid-configuration``
+  ``{issuer}/.well-known/openid-configuration`` (issuer is pinned)
+- ``expected_audience`` / ``expected_issuer``: pinned ``aud`` / ``iss`` claims;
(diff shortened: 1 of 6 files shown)


msaad00:
Folded in the UI bundle-budget headroom (3536→3552 KiB) so UI Validate passes — this PR touches `graph/effective_permissions.py`, which triggers UI Validate, and main was sitting at the exact bundle ceiling (the only failing step was `UI bundle budget`, unrelated to this identity change). Measured CI-variance headroom; largest-chunk and shared-runtime budgets unchanged.

andres-linero (approved):

agent-bom (approved):

PR #4184, opened 2026-07-18
feat(ciem): extend identity governance to AWS and GCP with real usage telemetry

msaad00:
## What
CIEM over-grant / dormant / orphan findings previously covered only Azure/discovered `managed_identity` nodes — AWS IAM roles (`ROLE`) and GCP service accounts (`SERVICE_ACCOUNT`) got no governance verdict at all. Separately, AWS Access Advisor usage evidence was collected but never consumed, so dormancy could not use real last-used telemetry.

## Changes
- `evaluate_identity_governance` now iterates `ROLE` and `SERVICE_ACCOUNT` in addition to `MANAGED_IDENTITY`, so over-grant/dormant findings span all three clouds. Provider attribution reads `provider` or `cloud_provider`.
- Builder threads collected AWS `usage_evidence` onto the identity node's `last_used_at` (newest real Access Advisor / RoleLastUsed timestamp), so dormancy uses real telemetry.

## Fail-closed (honesty)
For the cloud-inventory role/service-account types — which do not carry owner or last-used as a first-class contract — absent evidence is never turned into a verdict: no per-target usage signal ⇒ over-grant not evaluated; no last-used timestamp ⇒ not dormant; no `owner` key ⇒ not orphaned. `MANAGED_IDENTITY` behavior is unchanged.

Note: an AWS over-grant right-sizing `usage` map is intentionally **not** synthesized from Access Advisor — Access Advisor is service-namespace-level while grant targets are resource nodes, so a 1:1 mapping would risk false negatives. Over-grant for AWS roles stays fail-closed/not-evaluated until a resource-level usage source exists (wire point noted in `builder.py`). Dormancy, the primary telemetry goal, is fully wired.

## How verified
- New suite (8 tests, red→green): AWS role over-grant+dormant with `provider=aws`; GCP service-account dormant with `provider=gcp`; no-evidence role fabricates nothing; Azure behavior unchanged; `last_used_at` derivation unit + end-to-end.
- Governance regression suites `106 passed`; wider graph/builder/fusion `128 passed`; full-graph bootstrap `44 passed`. `ruff` + `mypy` clean.

### DIFF
--- a/src/agent_bom/graph/builder.py
+++ b/src/agent_bom/graph/builder.py
@@ -5585,6 +5585,33 @@ def _wire_network_entry_exposure_paths(
                     )
 
 
+def _role_last_used_at(usage_evidence: Any) -> str | None:
+    """Newest real last-accessed timestamp across role usage-evidence records.
+
+    Threads bounded AWS Access Advisor / RoleLastUsed telemetry
+    (:mod:`agent_bom.cloud.aws_iam_evidence`) onto the identity node so NHI
+    governance dormancy uses a real last-used signal. Returns ``None`` when no
+    record carries a timestamp — absent telemetry must never be turned into a
+    false "never used" (fail-closed); the role then stays not-evaluated for
+    dormancy rather than being fabricated as dormant.
+    """
+    if not isinstance(usage_evidence, Mapping):
+        return None
+    records = usage_evidence.get("records")
+    if not isinstance(records, list):
+        return None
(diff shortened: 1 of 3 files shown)


msaad00:
Combined into #4199 with #4187 (complementary CIEM work on graph/builder.py + nhi_governance.py). Merged there with both sides preserved; 330 nhi/ciem tests pass.

PR #4187, opened 2026-07-18
feat(ciem): surface over-privileged identities from AWS Access-Advisor usage evidence

msaad00:
## What

Closes the CIEM last-mile gap: `role["usage_evidence"]` (AWS Access-Advisor per-service last-used, collected since #4129) had **zero consumers** — no finding, no graph node, no surface. This wires it end-to-end so an identity granted permissions it never uses now produces an over-privilege finding.

## Wiring (usage_evidence → edge → finding)

- **Edge carrier** — `graph/builder.py` `_add_access_advisor_grants` emits one `HAS_PERMISSION` edge per granted service carrying Access-Advisor `last_used_at` (`None` = never used). **Honest counts:** edges are emitted *only* when `usage_evidence.state == "available"` (verified against the real `IamRoleUsageEvidence.to_dict()` schema); denied/pending/unavailable evidence yields no edges, so absence is never read as "unused".
- **Emitter** — `graph/nhi_governance.py` `build_ciem_over_privilege_findings` reads the Access-Advisor grant edges and flags services with no observed usage. Kept as a focused emitter (over-privilege only) rather than routing roles through the generic dormant/orphan path, which would fabricate unattended-identity findings for every ownerless role.
- **FindingType** — new `CIEM_OVER_PRIVILEGE` (`finding.py`), routed to the **cspm** security domain (`finding_scope.py`).
- **Report/surfaces** — folded into `AIBOMReport.to_findings()` (`models.py`), a standalone JSON block (`output/json_fmt.py`), emitted at graph-build (`cli/agents/scan_cmd.py`); `docs/DATA_MODEL.md` regenerated.

Scope: AWS Access-Advisor only — Azure/GCP usage right-sizing intentionally not built (smallest correct slice). Service-permission targets modeled as `RESOURCE` nodes (`kind=iam_service_permission`); no dedicated permission entity type exists yet.

## How verified

- `tests/test_ciem_over_privilege.py` (9): repro (unused service → finding of the new type naming the permission) + negatives (all-used → none; grants without Access-Advisor marker → none; `state != available` → no edges/no finding) + SARIF/domain/report-rehydration.
- I independently traced the field names + enum value (`UsageEvidenceState.AVAILABLE == "available"`, `to_dict` emits `state`/`records`/`service_namespace`) against the real producer to confirm the bridge fires in production, not just in tests.
- Touched suites (nhi_governance, aws_iam collector/evidence/usage, graph builder, effective-permissions, authorization-evidence, SARIF, finding_scope, models/json) — **1086 passed, 3 skipped**; `check-counts` + `check_release_consistency` green; `ruff` clean.

Note: `json_fmt.py`/`scan_cmd.py` picked up pinned `ruff-format` line-length normalizations on lines already touched — pure formatting.


msaad00:
Combined into #4199 with #4184 (both build CIEM governance on the same graph files; complementary). Merged there, both suites green on the combined result.

PR #4188, opened 2026-07-18
fix(ci): restore UI bundle budget headroom (unblocks UI Validate on backend PRs)

msaad00:
The Linux CI build measures a few dozen bytes **over** the total-client-JS ceiling (`FAIL total client JS: 3536.0 KiB / budget 3536.0 KiB` — both round to 3536.0, so the real overage is <60 bytes). It's measured runtime/bundler variance sitting exactly on a zero-headroom line, **not** a new feature chunk: `largest chunk 374/927 KiB` and `shared app runtime 417/439 KiB` both still pass with margin.

Because the total budget had no headroom, UI Validate (which the path classifier runs whenever a PR touches `src/agent_bom/api/routes/`) fails on **backend-only** PRs that don't change a byte of UI — e.g. #4180 (event-loop offload) and the exec-honesty PR. This restores ~16 KiB of headroom (3536 → 3552 KiB); largest-chunk and shared-runtime budgets are unchanged.

No bundle contents change — this is a one-line budget-constant bump with a justification comment, matching the file's established convention.

### DIFF
--- a/ui/scripts/check-bundle-budget.mjs
+++ b/ui/scripts/check-bundle-budget.mjs
@@ -48,7 +48,13 @@ const BUDGETS = {
   // investigation-lens controls. The combined Linux build measured 3511 KiB
   // (+23 KiB / 0.7% over the prior ceiling); 3536 KiB keeps bounded headroom
   // without changing the largest-chunk or shared-runtime budgets.
-  totalClientJsBytes: 3_620_864,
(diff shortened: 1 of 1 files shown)


msaad00:
Folded the one-line budget headroom directly into the three PRs that trigger UI Validate (#4180, #4187, #4189) so each is self-sufficiently green without a merge-order dependency on this PR. Identical one-line change, merges cleanly across all three.

PR #4193, opened 2026-07-18
fix(hub): atomic local ingest, bulk field validation, honest evidence truncation

msaad00:
Three write-path/ingest honesty defects.

## 1. SQLite/in-memory bulk ingest was non-atomic (P2)
Ledger append, current-state upsert, and reconcile committed as three separate transactions on the local backends (the atomic seam existed only on Postgres). A crash between the ledger commit and the current-state upsert left `tenant_total` inflated while the findings never appeared in any list.

**Fix:** implemented `ingest_batch_atomic` for `SQLiteComplianceHubStore` (single transaction, rollback on failure, cached ordinal/total advance only post-commit) and `InMemoryComplianceHubStore` (single lock, snapshot-restore on failure), matching the Postgres seam's `(new_total, reconciled)` contract. `hub_ingest.py` already prefers the atomic path, so all backends now use it.

## 2. Bulk ingest accepted garbage field types with 201 (P3)
`_normalized_bulk_finding` did no type validation, so `{"cvss_score": "NaNstring"}` or `{"severity": {"nested": true}}` returned 201 and materialised rows that leaked the value verbatim and could never match the severity/cvss filter.

**Fix:** severity must be a string (normalised to the canonical enum; unknown labels → `unknown` explicitly), cvss_score must be a finite number in 0.0–10.0 or null — otherwise a clear 422. Fails closed instead of storing a misleading row.

## 3. Signed compliance evidence bundle silently truncated at 10k (P3)
The audit-evidence fetch capped at 10,000 entries, then reported counts over the capped set with no marker — presenting partial evidence as the window's complete record.

**Fix:** when the fetch hits the cap, the signed body now carries `scope.audit_events_truncated` + `scope.audit_event_limit` and `audit_log_integrity.truncated`. Fixed the sibling `/compliance/report/pack` route too; the signature covers the flag.

## How verified
TDD red→green per defect (16 new tests incl. rollback-on-failure for both local backends, 422s for each bad type, truncation-marked vs not). Consolidated regression `75 passed, 7 skipped` (Postgres-only skips); broader bulk-route `181 passed`. OpenAPI regenerated; artifact + client-parity guards pass. `ruff`/`mypy` clean.

### DIFF
--- a/docs/openapi/v1.json
+++ b/docs/openapi/v1.json
@@ -1299,6 +1299,11 @@
             "title": "Tampered",
             "type": "integer"
           },
+          "truncated": {
+            "default": false,
+            "title": "Truncated",
+            "type": "boolean"
+          },
           "verified": {
             "default": 0,
             "title": "Verified",
@@ -1590,6 +1595,16 @@
             "default": 0,
             "title": "Audit Event Count",
             "type": "integer"
+          },
+          "audit_event_limit": {
+            "default": 0,
+            "title": "Audit Event Limit",
+            "type": "integer"
+          },
+          "audit_events_truncated": {
+            "default": false,
+            "title": "Audit Events Truncated",
+            "type": "boolean"
           },
           "completed_scan_count": {
             "default": 0,
--- a/docs/openapi/v1.yaml
+++ b/docs/openapi/v1.yaml
@@ -939,6 +939,10 @@ components:
           default: 0
           title: Tampered
           type: integer
+        truncated:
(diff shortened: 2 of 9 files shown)


msaad00:
Combined into #4198 with #4194 — the two conflicted on compliance_hub_store.py (both added methods at the same spot). Merged there with both sides' logic preserved; 214 tests pass + OpenAPI drift gate green on the combined result.

PR #4194, opened 2026-07-18
fix(read-path): bounded scope filtering and lifecycle-aware findings reads

msaad00:
Combined read-path correctness + honesty fixes (share the same query methods, so landed together to avoid a self-conflict).

## 1. Scope filter materialised the whole tenant table (P1)
Any `?provider`/`?account`/`?environment`/`?domain` dropped off the keyset path into an unbounded `SELECT … ORDER BY ordinal` + full Python sort — 1M rows loaded to return 50. Because these fields live in the JSON payload (`domain` is a computed overlapping-lens set), they can't be a single SQL predicate.

**Fix:** a bounded batched keyset scan (`collect_scope_filtered_page`) fetches ordered batches via the existing keyset query, applies the predicate to pre-enrichment rows, and collects only `limit`+1 matches — bounded memory, never materialises the table. The scope predicate has one source of truth (`finding_scope.row_matches_scope`), shared by route and store, so SQL and Python never diverge. Under a scope filter `total` is approximate (no O(table) COUNT); the unfiltered fast path is byte-for-byte unchanged.

## 2. Cursor pagination silently no-opped under a scope filter (P2)
The scope branch ignored `cursor` and never emitted `next_cursor`. Now the scope path rides the keyset cursor and emits `next_cursor` from the last emitted match (0-dup / 0-drop by construction).

## 3. `current_scan_findings` full-fold ran on every request (P2)
Recomputed on cursor/deep pages that discard it. Now skipped on cursor pages (scan findings only appear on page 1).

## 4. Finding lifecycle was invisible to every read surface (P1)
After a full-sync ingest resolved ~97k of 100k findings, `/v1/overview` still reported "grade F, 25,057 critical" and `/v1/findings` ranked resolved rows identically to open — there was no `status` filter anywhere.

**Fix:** `status` is a real sargable column on `hub_findings_current`. Added a `status` predicate to `list_current_page` + `current_severity_breakdown` (both backends) via a shared `status_sql_predicate`/`status_matches`, plus a partial open-findings index. The `/v1/findings` route gets a validated `status` enum param (open/resolved/all, default open; 422 on invalid), and the exec headline + severity strip now derive from the open-only basis — so resolved findings no longer inflate the grade. Live posture = `status IN ('open','reopened')`.

## How verified
TDD red→green: default view excludes resolved; `status=all`/`status=resolved` correct; overview `critical==3` after resolving 97/100; status+scope combined cursor paging 0-dup/0-drop with no resolved/scope leak; prior keyset/scope tests green. Both SQLite + in-memory verified end-to-end; Postgres predicate proven via generated-SQL tests (no live DB in dev). Suite `224 passed, 1 skipped`; `ruff`/`mypy` clean; OpenAPI regenerated, `check-counts`/schema/surface contract gates pass.


msaad00:
Combined into #4198 with #4193 (they conflicted on compliance_hub_store.py). Merged there, both sides' logic preserved, full gate green.

Issue #1969, opened 2026-04-26
typing: phase in stricter mypy settings for API and models


msaad00:
Audit tracker for Python typing depth.

Scope:
- Reduce Any and type: ignore use in api/ and models.py first.
- Enable stricter mypy settings module-by-module instead of globally flipping the repo.
- Keep GCP optional dependency import idioms documented or locally ignored with reasons.

msaad00:
Triage 2026-04-27. Iterative module-by-module work — #2003 (in v0.82.0 CHANGELOG) phased strict mypy onto four more API modules. The issue body says "module-by-module instead of globally flipping", so there's no single closing point. Each strict module reduces the open scope. Suggest closing only when the strict surface covers all of `src/agent_bom/api/` and the public model layer — currently partial. Issue stays open as a rolling tracker.

msaad00:
Per-module strict mypy now covers 13 modules (see `[[tool.mypy.overrides]]` in pyproject.toml: audit_log, compliance_signing, dashboard_csp, exception_store, fleet_store, idempotency_store, mcp_observation_store, metrics, policy_store, schedule_store, scim, scim_store, storage_schema, tenant_quota_store, tracing, backpressure, proxy_sandbox). Latest batch: #2031. Phasing continues; keeping open until the API tree is fully strict.

msaad00:
Progress: about 50%. Several typing hardening phases have landed, but stricter mypy is not fully phased through API and models. Leave open for incremental strictness PRs.

PR #4195, opened 2026-07-18
chore(typing): ban implicit generics in models.py and findings_current

msaad00:
Part of #1969 — one bounded module-by-module slice of the typing tracker (not a close).

## What
- `src/agent_bom/models.py`: parameterized 65 bare-generic annotations (`dict`→`dict[str, Any]`, `list`→`list[Any]`, `list[dict]`→`list[dict[str, Any]]`, `Union[dict, list]`→parameterized) on serialized scan-result fields. These were implicit `dict[Any, Any]`/`list[Any]`; naming the shape makes the JSON contract explicit. Annotation-only, no runtime change.
- `src/agent_bom/api/findings_current.py`: removed the two explicit `Any` parameters via a structural `_ScanJobLike` Protocol that `ScanJob` satisfies; kept the justified boundary `Any` on the `iter_findings` callable (narrowing it is contravariantly unsound against the call sites).
- `pyproject.toml`: a second `[[tool.mypy.overrides]]` block scoped to exactly these two modules, adding `disallow_any_generics` (one notch above the existing strict block) plus the standard strict flags. The base `[tool.mypy]` is untouched — no global flip, per the issue's module-by-module plan.

## How verified
Repo-wide CI mypy command (which reads the new override): `Success: no issues found in 768 source files` — no cascade into the call sites. `ruff` clean; touched suites `6 passed` + `91 passed`. Behavior-preserving typing change.

Part of #4207.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -445,6 +445,22 @@ disallow_incomplete_defs = true
 warn_return_any = true
 warn_unused_ignores = true
 
+# Phase 5 (#1969): tighten one strictness notch further — ban bare/implicit
+# generics (`dict`/`list` -> `dict[str, Any]`/`list[Any]`) so serialized-payload
+# fields declare their shape instead of degrading to `Any`. Applied
+# module-by-module; expand this list as each module is de-genericized. Not a
+# global flip — the base [tool.mypy] settings stay unchanged.
+[[tool.mypy.overrides]]
+module = [
+    "agent_bom.models",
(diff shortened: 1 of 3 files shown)


msaad00:
Combined into #4206 with #4200 (they conflicted on pyproject.toml — both added the strict disallow_any_generics override block). Merged there with the module lists unioned; mypy passes on all 7 strict modules.

PR #4201, opened 2026-07-18
feat(mitre): typed ATT&CK/ATLAS technique mappings on multi-hop attack paths

msaad00:
Part of #4108 — the "typed attack-path MITRE enrichment" workstream (core). The freshness workstream is #4197; the investigation-UI is a deferred follow-up. After this + the UI PR, #4108 closes.

## What
- **Model** (`graph/container.py`): `TechniqueMapping` (`hop_index`, `technique_id`, `technique_name`, `catalog`, `tactics`, `provenance`, `confidence`) + `AttackPath.technique_mappings`, dict round-trip, `mitre_technique_ids()`, tolerant JSON decoder.
- **Derivation from evidence, not label text** (`graph/attack_path_mitre.py`): each hop maps from the observed edge relationship + target entity type + edge evidence dict — never from `summary`/edge-label strings. Unmappable hops stay unmapped (fail-closed). Every ID validates against the bundled ATT&CK + ATLAS catalogs; tactics read from the catalog (no dangling IDs). Provenance is evidence-derived and never claims detected activity (it describes observed graph structure).
- **Wiring**: builder populates mappings after fusion; route-derived governance/toxic paths enriched on read so the API is fully populated.
- **Persistence** (symmetric SQLite + Postgres): new `technique_mappings` column + migration + read projections in all three store layers.
- **Contract**: OpenAPI schema extended + regenerated; graph JSON export carries the fields via `to_dict()`. (No SDK model to regen; SARIF/CDX/SPDX serialize paths at the finding level only, left unchanged.)

## Deferred
Investigation UI — a pure render of `technique_mappings[]` + `mitre_technique_ids` from `/v1/graph/attack-paths` (both already populated).

## How verified
9 new tests red→green (evidence-ordered mappings, exploit-public-facing from internet entry, ATLAS tool-reach, unmapped-when-no-evidence, never-claims-activity, dict/SQLite round-trip, tenant isolation) + fake-conn Postgres persist→load; broad graph sweep `453 passed, 10 skipped`; `ruff`/`mypy` clean; OpenAPI `--check`, `check-counts`, product-surface + freshness gates pass.

Note: best merged after #4197 (v18.1 catalog) so tactic labels resolve against the refreshed catalog; functions independently since tactics always resolve from whatever catalog is bundled.

Closes #4203.

PR #4210, opened 2026-07-18
fix(graph): consume admin-equivalent, de-prioritize mitigated exposure, surface skip reason

msaad00:
Part of the "computed but never surfaced" defect class — graph/posture signals stamped on nodes but read by nothing. Consumption-side only (no `builder.py`/`pipeline.py` changes).

## Fixes
1. **`admin_equivalent` was dropped (HIGH, CIEM)** — stamped by `effective_permissions` but read by nothing, while its sibling `escalates_to_admin` is consumed. Now surfaced as a fusion signal (with `admin_equivalence_basis`), a node boost, an `admin_reachable` toxic factor, and a standalone CIEM path so a principal holding admin directly appears in `/v1/graph/attack-paths`.
2. **CNAPP mitigation flags were dropped → over-report (HIGH, honesty)** — `protected_by_waf`/`exposure_mitigated`/`toxic_exposed_vulnerable_mitigated` were never read, so a WAF-fronted node ranked at full toxic-exposure weight. Now de-prioritized in fusion signals, node boost, and the toxic band — and in every case **visibly marked mitigated** (signal kind, path prefix, factor label), never silently dropped. `sensitive_data_access_count` also surfaced.
3. **"Capped, skipped-not-clean" signal was discarded (MED, honesty)** — `apply_effective_permissions` now persists `GraphAnalysisStatus(SKIPPED, reason_codes=("principal_cap_exceeded",))` on the graph (COMPLETE otherwise), so a large estate reads as "not computed", never "clean".

Not touched (honestly deferred): the `usage=` over-grant map (service-namespace vs resource-node mismatch), redundant `nhi_risk_*` attrs, cost_overlay attrs.

## How verified
9 new tests red→green; touched suites `155 passed`, regression net `147 passed`; `ruff`/`mypy` clean; OpenAPI `--check` OK (new keys ride existing `additionalProperties`). Adversarial reconciliation on a seeded twin-EC2 graph: bare toxic 82.0 vs WAF-mitigated 65.0 "(exposure mitigated)", admin-equivalent role 78.0; overlay counts reconcile, elif ordering can't double-count.

### DIFF
--- a/src/agent_bom/api/routes/graph.py
+++ b/src/agent_bom/api/routes/graph.py
@@ -393,18 +393,59 @@ def _fusion_signals_for_path(graph: UnifiedGraph, hops: list[str]) -> list[tuple
             continue
         attrs = node.attributes
         port_detail = _exposed_port_detail(attrs)
+        # A WAF / API-gateway in front of the resource mitigates its exposure
+        # (set by the CNAPP overlay). Honesty (§11): a mitigated node is NOT
+        # counted as a full-weight toxic/exposed foothold, but it is NOT hidden
+        # either — it surfaces as an explicitly *mitigated* signal at lower rank.
+        exposure_mitigated = bool(attrs.get("exposure_mitigated") or attrs.get("protected_by_waf"))
         if attrs.get("toxic_exposed_vulnerable"):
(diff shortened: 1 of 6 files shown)


msaad00:
Superseded by #4213, which combines this with #4201 (same graph layer, overlapping files) — all content and verification carried over. Branch left intact.

PR #4211, opened 2026-07-18
fix(compliance): wire CIS benchmark posture into the scorecard and narrow the framework over-claim

msaad00:
Part of the "computed but never surfaced" defect class. **⚠️ Changes the compliance scorecard — please review the numbers closely.**

## Root cause
`/v1/compliance` built every control status only from `blast_radius` CVE tags and never read `report.cis_benchmark_data` — so a cloud account **failing CIS Foundations Benchmark controls showed a green/`no_data` scorecard** even though `/cis/checks` reported the failures.

## Fixes
1. **CIS benchmark → scorecard**: new `cis_foundations_benchmark` line built from the **same** `build_cis_benchmark_check_rows` + same latest-per-`(cloud,check_id)` dedup as `/v1/cis/checks`, so the two reconcile by construction. Its FAIL/pass/error fold into `overall_status`/`overall_score` — a failing-CIS account now reads `fail`.
2. **No double-count**: kept **distinct** from the CVE-driven CIS Controls v8 line (Foundations `CIS-2.1.1` vs safeguard `CIS-02.1` never share IDs).
3. **ERROR honesty**: explicit `error` bucket; denominator `evaluated = pass+fail+error`, `score = pass/evaluated` — ERROR never folds into pass/fail; an error-only account reads `warning`, not a clean pass (consistent with `finalize_read_coverage` refusing to PASS permission-denied controls).
4. **Narrowed the over-claim WITHOUT fabricating a crosswalk**: `CLOUD_CIS` previously fanned out to CIS+SOC2+ISO+NIST wholesale; now asserts **only CIS** (the framework it actually failed). `applicable_frameworks` consistency added to `cloud_cis_check_to_finding`.

## How verified
7 tests red→green incl. failing-CIS-not-green, error-not-a-pass, no-double-count, reconciles-with-/cis/checks. Suites `426 passed` / final CIS batch `119 passed, 5 skipped`; `ruff`/`mypy` clean; OpenAPI/schema/check-counts pass. Adversarial TestClient render: seeded failing-CIS account → `overall_status: fail`, `score 25.0`, counts match `/cis/checks` exactly.

## Deferred (blocked on data)
An authoritative CIS-Foundations → SOC2/ISO/NIST **specific-control crosswalk** needs licensed CIS mapping data not in the repo — cross-framework attribution is intentionally **withheld rather than invented** until it lands. Same-class follow-up flagged: the Databricks `CLOUD_SECURITY` source still carries a source-level SOC2/ISO/NIST baseline (next to narrow).

### DIFF
--- a/src/agent_bom/api/routes/compliance.py
+++ b/src/agent_bom/api/routes/compliance.py
@@ -322,6 +322,101 @@ def _evaluated_control_status(sev_breakdown: dict[str, int]) -> str:
     return "not_evaluated"
 
 
+def _aggregate_cis_foundations_checks(jobs: list[Any]) -> dict[str, Any]:
+    """Deduped CIS Foundations Benchmark status rollup across a tenant's scans.
+
+    Reuses ``build_cis_benchmark_check_rows`` + the same latest-per-(cloud,
+    check_id) dedup as ``/v1/cis/checks`` so the scorecard's benchmark line
+    reconciles with that endpoint by construction.
+
+    This is the CIS *Foundations Benchmark* taxonomy (``CIS-2.1.1`` …), a
+    different standard from the CVE-driven CIS Controls v8 line (safeguard
+    ``CIS-02.1`` …) built elsewhere in :func:`get_compliance`. The two never
+    share identifiers, so no control is counted in both lines.
+    """
+    from agent_bom.analytics_contract import build_cis_benchmark_check_rows
+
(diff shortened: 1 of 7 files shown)


msaad00:
Superseded by #4215, which combines this with the other two 'computed but never surfaced' fixes (disjoint file sets, one defect class) — all content and verification carried over. Branch left intact.

PR #4212, opened 2026-07-18
fix(api): emit CIEM, toxic-combination, and MCP-tool-rule findings on the hosted scan path

msaad00:
Part of the "computed but never surfaced" defect class. **⚠️ Changes finding counts on the API/hosted path — please review closely.** Root cause: the API/hosted scan emitted fewer finding categories than the CLI, because several evaluators stopped at the graph or CLI and never joined the unified `to_findings()` stream that exec/severity/compliance counts and every finding output derive from.

## Fixes (via a shared CLI+API helper `attach_graph_derived_findings`)
1. **NHI/CIEM governance findings (HIGH)** — `build_nhi_governance_findings` had no production caller; the builder now computes them on the shared path and routes them into the report. Forward-compatible with #4199 (AWS/GCP arrive through the same emitter).
2. **Toxic-combination findings (MED-HIGH)** — the API output phase now runs the evaluator (it was CLI-only).
3. **MCP tool-rule findings (MED)** — `schema_rule_findings` → `Finding` inside `to_findings()`.

Interaction-risks **deliberately not wired** (double-count risk with existing a2a/toxic rules; `compute_interaction_risks` isn't on the API path) — follow-up.

## Honesty/correctness guards
- **No double-count**: exec/severity totals come solely from `to_findings()` (id-deduped); graph attack-paths stay a separate metric. Tested.
- **Idempotency**: all three carry deterministic ids → stable across re-runs; hub ingest dedups. Tested.
- **Tenant isolation** + **fail-closed** (empty stays empty, never raises into the scan). Tested.
- **Exporter reality**: SARIF + JSON + `/v1/findings` + severity gate get all three; **CDX/SPDX are CVE/package SBOMs and by design carry only CVE findings** (unchanged).

## How verified
66 new tests + broad regression `1217 passed, 11 skipped` (mcp suites `148 passed` with the extra installed); `ruff`/`mypy` clean; CLI↔API parity asserted on the same fixture.

## Follow-ups flagged (same class)
- **`iac_findings_data`** reaches JSON + SARIF's dedicated IaC loop but **not `to_findings()`** → under-counts in exec `total_findings`, the severity gate, and CDX. Needs a scoped fix that avoids double-counting SARIF's IaC loop.
- API graph is built twice per scan (findings + persist) — a noted optimization at >5000-node scale, not a correctness issue.

### DIFF
--- a/src/agent_bom/api/pipeline.py
+++ b/src/agent_bom/api/pipeline.py
@@ -35,6 +35,7 @@ from agent_bom.api.stores import (
     _jobs_put,
 )
 from agent_bom.config import API_SCAN_WORKER_RECYCLE_JOBS, API_SCAN_WORKERS
+from agent_bom.graph.scan_findings import attach_graph_derived_findings
 from agent_bom.security import sanitize_error
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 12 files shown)


msaad00:
Superseded by #4215, which combines this with the other two 'computed but never surfaced' fixes (disjoint file sets, one defect class) — all content and verification carried over. Branch left intact.

PR #4225, opened 2026-07-18
chore(ci): path-conditional Python suite + fuzzing, 3 pytest workers on PRs

msaad00:
Follow-up to #4222 — the conditional half of the CI fast path.

## What
- **Path classifier** gains a `python` output — conservative by construction: anything under `src/`, `tests/`, `scripts/`, `data/`, `contracts/`, packaging, `.github/actions/`, or ci.yml itself counts as Python-relevant.
- **Test job** skips only on a positive not-Python classification (docs/UI/deploy-only PRs). Fail-closed: if the classifier errors or reports nothing, the suite runs. Skipped required checks satisfy branch protection, so no PR gets stuck.
- **Build Package** tolerates a legitimately skipped test but still blocks on a failed/cancelled one.
- **PR-Fuzzing** (`cflite-pr.yml`, not a required check) gets a workflow-level paths filter for the same set.
- **PR test runs use 3 pytest workers** instead of 2 — PRs now run a single matrix job per #4222, and 4 workers was the observed runner OOM threshold, so 3 is the measured safe step. Main keeps the existing 4/2/2 split.

## Effect
Code PRs: ~10 → ~4–5 min (worker bump shortens the dominant Test job). Docs/UI-only PRs: ~2 min (Lint + Version Alignment + Security Scan only on the Python side).

## How verified
YAML parses; `check_product_surface_contract.py` and `check_workflow_timeouts.py` pass locally; this PR touches ci.yml so the classifier marks it python=true and alpine=true — its own run exercises the new gating (expect Test (3.13) to RUN here, with 3 workers).

### DIFF
--- a/.github/workflows/cflite-pr.yml
+++ b/.github/workflows/cflite-pr.yml
@@ -5,6 +5,15 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+    # Fuzzing exercises Python parsers only — skip on docs/UI/deploy-only PRs.
+    # Not a required check, so workflow-level path filtering is safe here.
+    paths:
+      - 'src/**'
+      - 'tests/**'
+      - 'pyproject.toml'
+      - 'uv.lock'
+      - '.clusterfuzzlite/**'
+      - '.github/workflows/cflite-pr.yml'
   schedule:
     - cron: '0 4 * * 1'  # Weekly batch fuzzing — Monday 04:00 UTC
   workflow_dispatch:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,7 @@ jobs:
       docs: ${{ steps.classify.outputs.docs }}
       endpoint: ${{ steps.classify.outputs.endpoint }}
       postgres: ${{ steps.classify.outputs.postgres }}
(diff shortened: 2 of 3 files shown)


msaad00:
Added a second commit: `--dist loadscope` → `--dist worksteal`. Evidence: the suite runs single-threaded in ~4 min (Alpine job) but took ~10 min under xdist — loadscope was serializing a few giant modules onto one worker. Full suite validated locally under worksteal: **15,391 passed, 0 failures, 296s**. This PR's own Test (3.13) run is the CI-side proof — compare its duration to the ~10 min baseline.

msaad00:
Worksteal is REVERTED from this PR (loadscope stays). Repeated full-suite runs under worksteal fail nondeterministically — first local run 15,391 passed, second run 6 failed, CI run 1 failed (a different test) — the suite has a latent class of order-dependent tests that loadscope's module grouping masks. One instance is fixed here (`test_api_surface_0943` warm phase made hermetic), and the class is filed as an engineering task with worksteal as the goal: the prize is the suite in ~210s vs ~576s. This PR keeps the safe changes only: path-conditional suite/fuzzing + 3 workers on PRs.

agent-bom (approved):

andres-linero (approved):

PR #4232, opened 2026-07-18
fix(postgres): make migrations authoritative at runtime

msaad00:
## Summary

- move all Postgres runtime-store tables, indexes, RLS policies, grants, and component markers into one idempotent migration-owned schema
- make configured API stores validate schema markers read-only and run with the DML-only `agent_bom_app` role
- fix clean-volume bootstrap by keeping the migration owner administrative and removing unsafe self-demotion
- make `make secrets` work before hosted URL/CORS configuration exists

## Root cause

Postgres store constructors were the only authority for 32 tables and many indexes/RLS policies. Disabling their startup DDL without migrating those objects made clean deployments fail, while the baseline's attempt to demote its active bootstrap role could leave a fresh volume partially initialized.

## Verification

- `uv --cache-dir /tmp/agent-bom-uv-cache run pytest -q tests/test_hub_write_path_scale.py tests/test_postgres*.py tests/api/test_api_operator_policy.py` — 250 passed, 6 skipped
- `uv --cache-dir /tmp/agent-bom-uv-cache run ruff check ...` — passed
- `uv --cache-dir /tmp/agent-bom-uv-cache run mypy src/agent_bom` — 776 files passed
- `make preflight` — passed
- full randomized suite reached 15,536 passed / 24 skipped with one stale test mock; the mock was corrected and its complete affected lane passed above
- clean Postgres 16 Docker init — 67 public tables and 31 schema-component markers
- empty-database packaged Alembic upgrade — reached head `20260718_01`
- `agent_bom_app` live proof — NOSUPERUSER, NOBYPASSRLS, no schema CREATE; tenant-scoped INSERT/SELECT succeeded and `CREATE TABLE` was rejected
- current Docker image API startup and `/health` — HTTP 200 using the DML-only Postgres connection

## Notes

The runtime still retains an explicit bootstrap path only for isolated tests/development pools that do not configure a Postgres deployment URL. Production configuration through either `AGENT_BOM_POSTGRES_URL` or a Postgres `AGENT_BOM_DB` value is migration-authoritative and read-only at startup.


msaad00:
Root cause of the red **Postgres Integration Contract**: this branch removed init.sql's final block that strips SUPERUSER/BYPASSRLS from the connecting bootstrap role — the #3665 tenant-isolation fix (a superuser role ignores FORCE ROW LEVEL SECURITY, silently voiding tenant RLS). The contract's bare `test ... = "f"` assertion fails with no output, which is why the job looked silent.

Restored the block. The stated removal concern (demoting the active session could break clean-volume init) doesn't reproduce: the block runs last and only affects the next connection. Verified the full contract sequence locally against Postgres 16 — init.sql lane: strip NOTICE fires, stamp + 14 migrations run on the demoted role; alembic-only lane: 16 upgrades + idempotent rerun, role ends non-super. If there's a concrete clean-volume failure case, let's reproduce it and handle it explicitly rather than dropping the strip.

msaad00:
Two more fixes pushed after full local repro against Postgres 16:

1. **The real clean-volume failure is now handled properly**: the baseline replay runs init.sql as the protected `postgres` bootstrap role (oid 10), which Postgres refuses to demote — that's why the strip looked broken. init.sql now strips every demotable role (compose/quickstart path keeps #3665 protection) and warn-skips only the undemotable bootstrap role. Verified: psql lane strips + migrates on the demoted role; bootstrap-role replay completes with a WARNING instead of aborting.
2. **CI now migrates `agentbom_test` before the integration tests** — with migrations authoritative, the test DB is a deployment like any other.

**Remaining author work (real divergence, not CI plumbing):** 3 integration tests fail with `ForeignKeyViolation` on `scan_jobs` — the migrated (init.sql-derived) schema enforces FKs to `teams` that the runtime write path never satisfies (the runtime-created schema didn't have them). Repro: migrate an empty DB, run `tests/test_postgres_integration.py` as `agent_bom_app`. Making migrations authoritative means resolving this schema contract: either the FK goes (migration) or the runtime creates team rows. That's a data-model decision, so leaving it to this PR's author.

msaad00:
Final piece pushed: the FK divergence is resolved by **provisioning the tenant's team row on first scan-job write** (idempotent `ON CONFLICT DO NOTHING` in the same transaction) rather than dropping the constraint — `teams` stays the FK root and referential integrity holds for dynamically-created tenants. Full contract now passes locally end-to-end on a fresh Postgres 16: empty DB migrated as the bootstrap role (16 upgrade steps, strip-or-warn behaves), then `tests/test_postgres_integration.py` **5 passed** as the non-superuser `agent_bom_app` role — including the cross-tenant RLS and cis-dedupe tests that were red. With the earlier commits (restored #3665 strip with bootstrap-role warn-skip, CI migrating `agentbom_test`), every finding from the original failure is addressed.

andres-linero (dismissed):

agent-bom (dismissed):

agent-bom (approved):

andres-linero (approved):

PR #4241, opened 2026-07-19
fix(delivery): redact webhook secrets and bound retries

msaad00:
## Summary

- redact secret-bearing webhook paths and queries in subscription/outbox API responses and ad-hoc sender logs
- validate retry timing and attempts with finite hard ceilings, and report actual attempts for permanent failures
- state the current outbox contract accurately: events are queued durably, while delivery requires an operator-run worker invoking `deliver_due_webhooks`

## Verification

- `UV_CACHE_DIR=/tmp/agent-bom-uv-cache uv run pytest -q tests/test_watch.py tests/test_watch_cov.py tests/test_watch_extra.py tests/test_governance_webhooks.py tests/test_posture_streaming.py tests/test_posture_webhook_outbox_api.py tests/test_proxy_webhook_retention.py tests/test_security_fixes.py tests/test_delivery_foundation.py tests/api/test_api_proxy_scorecard.py` — 156 passed, 2 skipped
- `UV_CACHE_DIR=/tmp/agent-bom-uv-cache uv run ruff check` on all changed Python and test files
- `UV_CACHE_DIR=/tmp/agent-bom-uv-cache uv run mypy` on all five changed source modules
- `UV_CACHE_DIR=/tmp/agent-bom-uv-cache uv run python scripts/check_exception_sanitization.py`
- `UV_CACHE_DIR=/tmp/agent-bom-uv-cache make preflight`
- `git diff --check`

## Notes

Fail-closed configuration behavior: invalid, non-finite, or excessive retry settings are rejected at construction. Delivery failures remain best-effort and dead-letter without raising into the caller.

PR #4242, opened 2026-07-19
fix(gateway): pin and validate upstream destinations

msaad00:
## Summary
- pin gateway relay and OAuth sockets to DNS answers validated at connect time
- keep control-plane-discovered upstreams public-only while operator-authored YAML approves private/VPC destinations
- deny cloud metadata and unsafe address classes in every mode; disable proxy-env routing and redirects

## Verification
- `uv run pytest -q tests/test_gateway_egress_transport.py tests/test_gateway_upstreams.py tests/test_gateway_server.py` (88 passed)
- `uv run ruff check src/agent_bom/runtime/egress_transport.py src/agent_bom/gateway_upstreams.py src/agent_bom/gateway_server.py tests/test_gateway_egress_transport.py tests/test_gateway_upstreams.py tests/test_gateway_server.py`
- `uv run mypy src/agent_bom/runtime/egress_transport.py src/agent_bom/gateway_upstreams.py src/agent_bom/gateway_server.py`
- `uv lock --check`
- `uv run python scripts/check_package_layout.py`
- `uv run python scripts/check_exception_sanitization.py`
- live loopback JSON-RPC relay smoke through the pinned private transport
- `git diff --check`

## Notes
- HTTP Host, TLS SNI, and certificate hostname verification continue to use the configured hostname; only the socket destination is pinned.
- Local YAML is the trust boundary for legitimate in-cluster/private upstreams. Discovery payloads cannot forge that provenance.


msaad00:
Superseded by #4245, which combines this with #4241 (both outbound network-boundary hardening, disjoint files) — content and verification carried over. Branch left intact.

Issue #4274, opened 2026-07-20
Epic: consolidate auth to a single principal-resolution pipeline (ordered resolvers, posture matrix, residual closures)


msaad00:
## Why

The API auth surface has grown to ~4,800 lines across `api/middleware.py`, `api/auth.py`, `api/oidc.py`, `api/oidc_browser.py`, `cli/_server.py`, plus 14 auth-adjacent modules (SCIM bearer, browser sessions, OAuth AS, Snowflake OAuth, model-key broker, trusted-proxy). Each credential path is individually fail-closed (verified hands-on 2026-07-19 at 3e8a41841), but the paths are parallel: resolution logic lives inside `APIKeyMiddleware` as a de-facto chain (`_try_scim_bearer_auth` → `_try_browser_session_auth` → `_try_proxy_header_auth` → key check → anonymous fallback), and posture reporting (`get_auth_runtime_status`, `trusted_proxy_auth_usable`, CLI serve gate) re-derives the same facts in several places. #4271 fixed exactly this class of bug (posture said configured, middleware disagreed). The combinatorics of N interacting modes can't be exhaustively covered by audits; they can be covered by structure + a contract-test matrix.

## Goal

One principal-resolution pipeline; every auth mode becomes a credential *source*, not a parallel path. Authorization (RBAC/tenant) happens once, after resolution.

```
resolve_principal(request) -> Resolved(principal) | Invalid(reason) | Absent
# ordered resolver chain, first non-Absent wins:
#   1. OIDC bearer (incl. SCIM bearer)
#   2. API key
#   3. Browser session
#   4. Trusted-proxy headers (external-auth escape hatch: auth_proxy / oauth2-proxy / Cloudflare Access pattern)
#   5. Anonymous (opt-in role assignment for unresolved principals — not a mode)
# Invariant (structural, tested): Invalid NEVER falls through to the next resolver
# — a presented-but-invalid credential is a hard 401, never a downgrade.
```

## Workstream (small PRs, in order)

1. **Extract the resolver chain** from `APIKeyMiddleware` into explicit ordered resolvers returning `Resolved | Invalid | Absent`; single policy point after resolution. No behavior change; golden tests before/after.
[...]

PR #4304, opened 2026-07-20
refactor(auth): posture single-source-of-truth + decide the OIDC-invalid fall-through

msaad00:
Part of #4274 (PR2).

## What consolidated

The auth posture was derived in several places (`get_auth_runtime_status`, `get_trusted_proxy_auth_status`, `trusted_proxy_auth_usable`, the CLI serve gate, and configure_api each re-deriving the same env facts). This PR introduces **one derived `AuthPosture`** as the single source of truth and wires all readers to it.

- **`AuthPosture` + `derive_auth_posture()`** (`api/middleware.py`) — the one place the effective credential **sources (in resolver precedence order)**, **anonymous on/off**, **listener scope (loopback / non-loopback)**, **trusted-proxy usability**, and **OIDC status** are derived. `apply_auth_posture()` stores it and projects the operator/UI runtime status; `get_auth_posture()` exposes it.
- **Readers aligned, no re-derivation:**
  - `configure_api()` derives once and applies it → middleware runtime status.
  - `/health` (and operator surfaces) read `get_auth_runtime_status()`, projected from the same posture.
  - the CLI **serve/api gate** (`_enforce_auth_defaults`) reads the same posture instead of hand-deriving `has_api_key/has_oidc/has_scim/has_saml/has_trusted_proxy`.
  - `configure_auth_runtime()` is kept as a backward-compatible entry point that builds + applies a posture, so `configured_modes` / `recommended_ui_mode` and the SSO tests keep their exact contract.
- **One structured startup line** summarizing the posture (sources + precedence + anonymous state + listener scope + trusted-proxy/OIDC status).
- **Contradictory combo → fail-fast startup error** via `validate_auth_posture()`: a malformed/contradictory OIDC configuration (which today only 500s on the first OIDC request) now fails at the serve/api boot gate. **Consolidation, not a policy change** — every combination that boots and serves today keeps working, including **trusted-proxy alongside direct OIDC on one listener** (explicitly tested), and the programmatic serve-gate set is unchanged (browser-only SSO still does not satisfy a non-loopback gate).

## The OIDC-invalid decision (rationale)

PR1 flagged that an OIDC decode/verify failure fell through to the API-key store as `Absent`. Decision made here:

- A **JWT-shaped** bearer (`header.payload.signature` with an `alg` header) that fails OIDC verification — bad signature/issuer/audience, expired, unconfigured issuer, replay, malformed claims — **was presented as** an OIDC token. It is now terminal **`Invalid` (hard 401)** and is **never retried against the API-key store**.
- A bearer that is **not JWT-shaped** (a raw opaque API key sent as `Authorization: Bearer`) stays **`Absent`**, so the legitimate **key-as-bearer** path still authenticates.

`token_is_jwt_shaped()` (`api/oidc.py`) is a structure-only discriminator — no signature/claim verification, never treated as trust. Rationale: a presented OIDC credential that cannot be verified must fail closed at its own resolver, not silently downgrade to a key attempt. (Note: an unresolvable **configured** issuer still surfaces as a 500 from the SSRF guard's `SecurityError` — pre-existing operator-misconfig behavior, driven by the configured issuer not the token, and out of PR2 scope.)

PR1 invariants preserved: `Invalid` never falls through; anonymous opt-in VIEWER-clamp; loopback dev-key.

## Tests (TDD)

- `tests/api/test_auth_oidc_invalid_decision.py` — JWT-shape unit matrix; invalid OIDC bearer → 401 and **never offered to the key store** (spy); raw key as Bearer → 200; OIDC success path unchanged.
- `tests/api/test_auth_posture_sot.py` — derivation facts + precedence; SoT alignment (runtime status derived from the applied posture); invalid-OIDC startup error; valid trusted-proxy+OIDC combo stays allowed; structured startup line.
- `tests/test_cli_serve_auth_enforce.py` — serve gate reads the posture (browser-only SSO still refuses; contradictory OIDC config fails fast).

## Verification

- `ruff check` + `ruff format --check` + `mypy` clean on changed files.
- Full `tests/api/` + CLI serve suites: **942 passed**. Named auth suites all green (hardening, oidc, oidc_browser, resolve_principal_pipeline, operator_policy, snowflake_oauth, sso_provider_presets, auth_debug, scim_lifecycle, no_ui_and_auth_posture, cli_serve_auth_enforce, cli_server, auth, shared_auth_state, proxy_audit, gateway_auth, mcp_auth_posture).
- Live app smoke through lifespan: startup posture line emitted once; invalid JWT → 401; static key as Bearer → 200.

Not in scope (later PRs): contract-test matrix (PR3), key↔principal-ID binding (PR4).


msaad00:
Superseded by #4309 (combined auth posture/OIDC + audit-checkpoint backfill into one security PR per the minimal-PR request; #4305's test-mock bug fixed there). Branch retained on origin; no work lost.

Issue #4284, opened 2026-07-20
P1(audit): chain-head read orders by entry-creation timestamp — a retried append poisons the head and livelocks all future appends for the tenant


msaad00:
Found while regression-testing #4276 on live Postgres (pre-existing on main @3e8a41841 — reproduced with the #4276 fix stashed, so independent of it).

**Symptom:** `tests/test_audit_chain_concurrency.py::test_postgres_concurrent_append_does_not_fork_chain` fails against a live Postgres (fresh Alembic-migrated DB, container PG 5433): with 8-16 barrier-aligned writers, several appends exhaust all `_MAX_APPEND_RETRIES` (50) retries on the fork-guard unique index and raise.

**Root cause:** `AuditEntry.timestamp` is assigned at **entry creation**, but `PostgresAuditLog._latest_signature_for_tenant` (src/agent_bom/api/postgres_audit.py) selects the chain head by `ORDER BY timestamp DESC, entry_id DESC`. A writer that loses the fork race re-signs against the advanced head and commits **later** with its **original (older) timestamp**. The chain-link order then diverges from timestamp order, e.g. (observed rows, one tenant):

```
timestamp                          eid       prev      sig
...827783                          aab44759            0cec7f04   <- genesis
...828025                          b234af14  650f9ab7  887eb0db   <- true head (committed last, older ts)
...828097                          0d8eb84e  0cec7f04  7e31285c
...828132                          6b70479e  7e31285c  650f9ab7   <- max(ts) = what the head read returns
```

The head read now returns `650f9ab7` (max timestamp) — but that head's successor slot is already taken by `b234af14`. Every subsequent append for the tenant signs against the stale head, hits the fork-guard UNIQUE violation, re-reads the same stale head, and retries until exhaustion: a **permanent livelock** — the tenant can never append again, and `verify_integrity`'s chronological walk mis-scores the out-of-order chain.

**Fix direction:** make head selection chain-aware rather than timestamp-ordered — e.g. read the head from `audit_chain_checkpoint.head_signature` (already bumped transactionally with each insert on the same connection), or select the row whose `hmac_signature` appears in no other row's `prev_signature` for the tenant; and/or restamp `entry.timestamp` on each retry so commit order and timestamp order agree.

**Severity/target:** P1 (audit availability — a single retried write can permanently wedge a tenant's audit trail); next patch release.

[...]

PR #4305, opened 2026-07-20
fix(audit): backfill legacy audit-chain checkpoints so truncation checks are correct

### LINKED ISSUE #4294: audit: one-time checkpoint backfill for legacy tenants (entry_count under-seeding)
Follow-up flagged during #4284 (PR #4293).

For a **legacy** tenant whose `audit_log` rows predate its first `audit_chain_checkpoint` upsert, `_upsert_checkpoint` seeds `entry_count=1` rather than the true historical count. #4293's livelock fix (chain-order head read + successor-free fallback) makes appends self-heal, but `verify_integrity`'s truncation check can under-count for such tenants until enough new appends accrue.

Behavior is unchanged by #4293 (pre-existing edge, not a regression). Fix: a one-time checkpoint backfill migration that computes the true per-tenant `entry_count` and head from `audit_log` links. Low priority — only affects tenants with audit history predating the checkpoint table.

Path: `src/agent_bom/api/postgres_audit.py` (`_upsert_checkpoint`, `verify_integrity`).

msaad00:
## Problem

For a **legacy** tenant whose `audit_log` rows predate its first `audit_chain_checkpoint` upsert, `_upsert_checkpoint` seeded `entry_count=1` instead of the true historical count. In a migration-owned deployment the checkpoint table is created empty and the runtime store never runs `_hydrate_checkpoints` (Postgres authoritative), so a legacy tenant's first append seeds the checkpoint from scratch. `verify_integrity`'s truncation check (`len(entries) == checkpoint.entry_count`) then **under-counts** for that tenant until enough new appends accrue — mis-scoring a valid intact chain and weakening tail-truncation detection. Pre-existing edge flagged during #4284/#4293, not a regression. `#4294`

## Fix

- **Seed from the true count.** The INSERT branch of `_upsert_checkpoint` (Postgres and SQLite) now sets `entry_count` from `SELECT COUNT(*) FROM audit_log` for the tenant, computed in the **same transaction** as the audit_log INSERT (so it includes the just-inserted row). The steady-state `ON CONFLICT` path stays an O(1) increment; a genuine genesis still seeds 1. This prevents recurrence for every deploy mode.
- **Idempotent reconciler.** New `backfill_checkpoints()` on both stores recomputes `entry_count` and the successor-free chain-tip `head_signature` from `audit_log` links for every tenant, healing already-under-seeded checkpoints. The Postgres path runs under the RLS-bypass maintenance context; head derivation mirrors `_chain_tip_signature_from_log` / #4293 (chain-order-correct under timestamp skew).
- **One-time migration** (`20260720_01`) runs the same reconciliation in SQL under `app.bypass_rls` for migration-owned deployments, where the app never batch-backfills. Idempotent; no meaningful downgrade.

Preserves the #4283 tenant-GUC binding, #4293 chain-order head logic, and the "audit must not block auth" append behavior. No new API/CLI/MCP field — the existing audit-verify surfaces consume `verify_integrity`'s `(verified, tampered)`, which becomes correct for legacy tenants.

## Test / verification (live Postgres, fresh Alembic-migrated DB, RLS on; plus SQLite)

New `tests/test_audit_checkpoint_backfill_4294.py`:
- An under-seeded checkpoint (`entry_count=1` while N rows exist) makes `verify_integrity` mis-score the intact chain **before**; after `backfill_checkpoints()` the checkpoint is `entry_count==N`, `head==true tip`, and `verify_integrity` is clean.
- **Idempotent** — a second run yields the same checkpoint.
- A **correctly-seeded** (non-legacy) tenant is left untouched.
- The append **seed** no longer bakes `entry_count=1` for a legacy tenant (`==N+1`).
- The migration is chained from the current head, bypasses FORCE RLS, and reconciles via `ON CONFLICT DO UPDATE`.

Verified live on the migration-owned schema: the migration healed a corrupted checkpoint `1 -> 7` as the non-superuser `agent_bom_app` role and is idempotent across downgrade+re-upgrade. The mock connection pool was updated to model the seed COUNT subquery. Audit chain/concurrency + `#4293` suites stay green; `ruff` + `ruff format` (pinned) + `mypy` clean on changed files.

Closes #4294

### DIFF
--- a/deploy/supabase/postgres/alembic/versions/20260720_01_audit_checkpoint_legacy_backfill.py
+++ b/deploy/supabase/postgres/alembic/versions/20260720_01_audit_checkpoint_legacy_backfill.py
@@ -1,0 +1,68 @@
+"""Backfill legacy audit-chain checkpoints from the audit_log links (#4294).
+
+A tenant whose ``audit_log`` rows predate its first ``audit_chain_checkpoint``
+upsert was seeded at ``entry_count=1`` rather than its true historical count:
+the migration-owned schema creates ``audit_chain_checkpoint`` empty and the
+runtime store never runs ``_hydrate_checkpoints`` on the authoritative path, so
+a legacy tenant's first append seeds the checkpoint from scratch. This one-time,
+idempotent reconciliation recomputes each tenant's ``entry_count`` (row count)
+and ``head_signature`` (the successor-free chain tip, chain-order-correct
+regardless of timestamp skew — #4284/#4293) directly from ``audit_log`` so
+``verify_integrity``'s truncation check stops under-counting for those tenants.
+
+The append seed path is fixed in the same change, so this only heals rows that
+were already under-seeded before the fix shipped; rerunning is a no-op.
+
+Revision ID: 20260720_01
+Revises: 20260719_03
+"""
+
+from __future__ import annotations
+
+from alembic import op
+
+revision = "20260720_01"
+down_revision = "20260719_03"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # audit_log (team_id) and audit_chain_checkpoint (tenant_id) both enforce
+    # FORCE ROW LEVEL SECURITY, so this cross-tenant maintenance write runs under
+    # the app.bypass_rls GUC the RLS policies honour (transaction-local). The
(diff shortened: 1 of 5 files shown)


msaad00:
Superseded by #4309 (combined auth posture/OIDC + audit-checkpoint backfill into one security PR per the minimal-PR request; #4305's test-mock bug fixed there). Branch retained on origin; no work lost.

PR #4307, opened 2026-07-20
fix(export): widen parquet/lake export to all unified finding types

### LINKED ISSUE #4280: P2(export): parquet export limited to CVE+malicious — widen to unified finding types (schema-versioned)
Sibling of the CSV completeness fix in #4279: `src/agent_bom/output/parquet_fmt.py:105` exports via `machine_export_findings`, which carries only CVE + malicious-package findings. COMBINATION / PROMPT_SECURITY (and future unified types) never reach the parquet/lake surface.

Not folded into #4279 because widening changes a typed schema consumed downstream (Iceberg/lake ingestion) — needs a deliberate schema-version bump or additive-column strategy rather than a drive-by. Reuse `unified_export_findings` from #4279 once the schema plan is set.

msaad00:
## Problem
`parquet_fmt.to_arrow_table` built rows from `machine_export_findings`, which carries only **CVE + malicious-package** findings. `COMBINATION` / `PROMPT_SECURITY` (and other unified/graph-derived types) never reached the Parquet file or the Iceberg lake table — even though JSON/SARIF/CSV all surface them after #4279. On `scan --demo --offline` the Parquet file held **16 rows for a 25-finding stream** (sibling of the CSV gap fixed in #4279).

## Fix — additive, schema-versioned
Reuse `unified_export_findings` (the same helper #4279 used for CSV) so parquet covers every finding type. The schema change is **strictly additive** so downstream Iceberg/lake consumers do not break:

- Every existing **v1 CVE column keeps its exact name, order, dtype, and value** — CVE rows are byte-identical to before.
- Three new **nullable** columns (`finding_type`, `finding_id`, `title`) are **appended at the END**, mirroring the CSV fix. CVE-specific columns are null for non-CVE rows and vice-versa.
- No schema-version mechanism existed for the lake table, so this **adds one**: `PARQUET_SCHEMA_VERSION = "2"` embedded as pyarrow schema key-value metadata (`agent_bom.parquet_schema_version`). v1 = the 28-col CVE+malicious table; **v2** = the additive unified widening. No breaking rename/retype.

The Iceberg catalog exporter shares `to_arrow_table`, so the lake table is aligned automatically (docstring + shape test updated 28→31 cols). The scheduled lake-export path (`agent_bom/export/runner.py`) reads hub finding-dicts — a separate stream already carrying all types — so it was never affected.

## Row counts (demo, `scan --demo --offline -f parquet`)
- **Before:** 16 rows — CVE 15 + MALICIOUS_PACKAGE 1
- **After:** 25 rows — + PROMPT_SECURITY 4 + COMBINATION 5 (now matches CSV/JSON/SARIF)

## Surfaces reconciled
Parquet file · Iceberg lake table · CLI `-f parquet`. Scheduled warehouse export already carried all types (verified independent path).

## Tests (TDD, red→green)
- `test_parquet_exports_all_unified_finding_types` — row count == unified count; COMBINATION/PROMPT_SECURITY present with `finding_type`.
- `test_parquet_cve_columns_byte_identical_to_cve_only_export` — every v1 column value + dtype identical between a mixed-estate CVE row and the same CVE exported alone.
- `test_parquet_schema_is_additive_and_versioned` — v1 columns preserved in order at the front, new columns appended + nullable, schema-version metadata = `2`.
- `test_iceberg_catalog` shape assertion updated to 31 cols (additive widening).
- Gates: `ruff` + `ruff format --check` clean, `mypy` clean on changed files, 155 tests across parquet/iceberg/output-formats/scheduled-lake/malicious/interop/reachability/scan-parity green.

Closes #4280

### DIFF
--- a/src/agent_bom/output/iceberg_catalog.py
+++ b/src/agent_bom/output/iceberg_catalog.py
@@ -108,9 +108,10 @@ def register_findings(
     *,
     catalog: Any | None = None,
 ) -> dict[str, Any]:
-    """Append the report's CVE findings to the Iceberg table as a new snapshot.
+    """Append the report's unified findings to the Iceberg table as a new snapshot.
 
-    Creates the namespace and table (matching the shared 28-col Parquet schema)
+    Creates the namespace and table (matching the shared unified Parquet schema —
+    the 28 v1 CVE columns plus the appended finding_type/finding_id/title, #4280)
     if they do not yet exist, then appends a data file and commits a snapshot.
     Returns a small summary dict for logging.
 
--- a/src/agent_bom/output/parquet_fmt.py
+++ b/src/agent_bom/output/parquet_fmt.py
@@ -14,12 +14,19 @@ from agent_bom.models import AIBOMReport, BlastRadius
 from agent_bom.output.finding_views import (
     evidence,
     is_package_malicious,
-    machine_export_findings,
     package_ecosystem,
     package_name,
     package_version,
     severity_value,
+    unified_export_findings,
 )
+
+# Lake/Parquet schema version. v1 = the CVE+malicious 28-column table; v2 adds
+# the three appended unified columns (``finding_type``/``finding_id``/``title``)
+# and widens the row set to every unified finding type (#4280). The bump is
+# purely additive: no v1 column is renamed, retyped, or reordered, so existing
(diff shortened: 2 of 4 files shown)


msaad00:
Superseded by #4308 (combined graph + parquet into one PR per the minimal-PR request). Branch retained on origin; no work lost.

PR #4306, opened 2026-07-20
feat(graph): store-backed live UnifiedGraph (proven, unwired)

msaad00:
## Summary

Adds `StoreBackedUnifiedGraph` — a container presenting the **exact** public `UnifiedGraph` interface backed by the #4295 indexed build workspace (SQLite temp DB or shared Postgres), keeping only a **bounded working set** in RAM: an LRU node cache with **dirty-tracking write-back** and **keyset-paged** adjacency queries. This is the store-backed live graph the epic needs to remove the builder's own in-RAM materialisation and realise the #4055 peak-RSS bound — shipped here **proven but unused**.

- `add_node` merge-union, `add_edge` dedup + bidirectional-reverse materialisation, `get_node` / `nodes` mapping view, `edges` / `adjacency` / `reverse_adjacency` views. Every traversal / filter / view algorithm is **inherited unchanged** from `UnifiedGraph` and runs against the store-backed views.
- New workspace read primitives on **both** backends over existing #4295 indexed columns (no schema change): `get_edge_payload`, `fetch_node_page` / `fetch_edge_page` (keyset-paged — each call is one closed query, so write-back can interleave between pages without colliding with an open server-side cursor, the Postgres hazard), and an `open_workspace_backend` factory shared with `open_graph_build_workspace`.

## Scope / alignment

**Default-off and unwired** — no builder, overlay, or persist caller constructs the container; the shipped path still materialises the in-RAM `UnifiedGraph`. Tests are the only callers. **No API / CLI / MCP / export / UI-UX / DB-migration / schema surface change** (the #4295 schema is reused). Only `build_workspace.py` is modified (read helpers + factory); the container and tests are new files.

## Interface-identity audit (every overlay + `attack_path_fusion` + `effective_permissions` + `nhi_governance`)

- **No item-assignment or delete on `graph.nodes` anywhere.** Access patterns found: `graph.nodes[id]`-then-mutate-`attributes` (builder + effective_permissions), `id in graph.nodes`, `graph.nodes.get(id)`, `graph.nodes.values()` / `.items()` / `sorted(graph.nodes)`, full-set iteration **with in-place mutation** (cnapp/governance/nhi/evidence/repo overlays), `for e in graph.edges` + `len(graph.edges)`, and `graph.adjacency.get(id, [])`. The store-backed views support **all** of these — so the overlays run unchanged.
[...]

Issue #4075, opened 2026-07-16
Wire a streaming graph producer through the persist pipeline to realize the #4055 write-path memory bound


msaad00:
## Context

#4074 landed the streamed-persistence plumbing (`save_graph_streaming` on the SQLite backend, `iter_graph_nodes`/`iter_graph_edges` read primitives, and a bounded reconciliation-set fix on both SQLite and Postgres). That PR realizes a **modest, real** reduction on the production write path (the retired-edge reconciliation set is now bounded by the *previous* snapshot size instead of a second O(edges) set of every incoming key), and it delivers the bounded streaming primitive verified in isolation.

It does **not** yet realize the full peak-RSS bound described in #4055. The production writer `api.pipeline._persist_graph_snapshot` still:

1. materializes the whole correlated graph via `build_unified_graph_from_report(...)`, then
2. loads the **previous** snapshot into a second full `UnifiedGraph` for delta-alert computation (`compute_delta_alerts`), so peak holds up to **two** materialized graphs, and
3. calls `save_graph(graph)` with the already-materialized graph.

`save_graph_streaming` / `iter_graph_*` have no production caller that passes a lazy producer — so the 68-91% micro-benchmark figure reflects the streaming primitive in isolation, not the shipped write path.

## Work

- Stream the correlation/graph builder so `_persist_graph_snapshot` can feed `save_graph_streaming` node/edge **generators** instead of a fully built `UnifiedGraph`.
- Replace the second full `UnifiedGraph` load of the previous snapshot (for delta alerts) with a bounded diff that streams the prior snapshot (e.g. via `iter_graph_edges` / the keyset read API) rather than materializing it — otherwise peak still holds a full prior graph.
- Extend the streaming persist path to the Postgres backend's node + node-search double write (the single-pass iterable was left non-trivial in #4074).
- Add a production-path peak-RSS regression test (not just the isolated primitive) that asserts the pipeline persist stays bounded as N grows.

Progresses #4055.

msaad00:
Current-main validation after PRs 4091 and 4097:

- The prior-snapshot/digest side is fixed, and supported writers now persist incoming node/edge iterables without recreating a second full prior graph.
- The producer is still materialized: api/pipeline.py builds a full UnifiedGraph before passing graph.nodes.values() and graph.edges to the writer. The container also retains adjacency, reverse-adjacency, and dedup indexes while correlation overlays require whole-graph queries.
[...]

Issue #4055, opened 2026-07-16
scale: public graph write path materialises whole UnifiedGraph in memory (~3.3GB RSS at 1M nodes)


msaad00:
## Summary

Constructing and persisting a graph through the public write path
(`UnifiedGraph` built in memory via `add_node`/`add_edge`, then
`SQLiteGraphStore.save_graph(graph)`) materialises the entire graph in Python
memory before it is written. At **1,000,000 nodes + ~1,000,000 edges** this
costs **~3.28 GB max RSS (~4.19 GB peak footprint)** and ~60 s wall on the
write path. It does not crash on a large host, but a container with a typical
2–4 GB memory limit would OOM. This bounds the "scales to millions" claim: the
limit is the in-memory materialisation, not the storage backend (every
backend's `save_graph` takes a fully-built `UnifiedGraph`).

## Repro (measured, read-only verification)

Build 1M-node `UnifiedGraph` via the public API, then `save_graph` on the
default SQLite backend, under `/usr/bin/time -l`:

```
[build] 1,000,000 nodes + 999,999 edges in 7.2s
[save]  save_graph -> 29.7s, db=733MB
[load]  load_graph -> 17.6s (also re-materialises the full graph in memory)
59.78s real  | maximum resident set size 3,283,664,896 (3.28 GB)
             | peak memory footprint     4,193,979,776 (4.19 GB)
```

`load_graph` re-materialises the whole graph too, so reads at this scale are
equally heavy.

## Impact

- Demo / single-node (in-process) path: a ~1M-node correlated graph needs
  ~3–4 GB; ~5M would need ~20 GB and OOM on any normal container.
- No streaming / incremental build+persist path — the graph object must fully
  exist in RAM before and during `save_graph`, independent of storage backend.

## Proposed direction

Batched/streamed graph persistence (chunked node/edge flush during build) and a
bounded/streaming `load_graph` so peak RSS is decoupled from graph size; or an
explicit documented ceiling + guard (mirroring the findings in-memory
sort-ceiling warning) that steers large graphs to a server-side backend.

## Severity / target

- Severity: **P2** (scale-honesty; does not crash the normal small-graph path;
  affects only very large single-node graphs).
- Target: **next minor** (not a patch-release blocker). Until fixed, avoid
  claiming multi-million-node graphs through the in-process construct→save path.

PR #4313, opened 2026-07-20
feat(graph): opt-in store-backed graph producer to bound build peak RSS

msaad00:
## What

Wires the graph producer `build_unified_graph_from_report` and its Phase-B overlays to run against a `StoreBackedUnifiedGraph` behind a **default-off** flag, realizing a large, measured reduction of the producer's peak RSS — the last-mile of #4055's write-path memory work. Previously #4308 landed the store-backed container proven but **unwired**; this connects it to the shipped persist path.

## Design (owner-approved)

- **New default-off flag `AGENT_BOM_GRAPH_STORE_BACKED_BUILD`.** Off ⇒ byte-identical in-RAM producer; no latency or behavior change for anyone not opting in.
- **Per-build throwaway private SQLite build workspace** — never the shared Postgres workspace tables, so no cross-tenant workspace data lives mid-build and the workspace-RLS question is moot. Context-managed: the workspace is dropped after build+persist, even on exception.
- **Scoped to the persist caller** (`api.pipeline._persist_graph_snapshot`) via a new `build_unified_graph_from_report(container=...)` seam. CLI/API/export keep the in-RAM path.

## Overlays run unchanged — with two minimal store-safety fixes

The #4306 interface-identity audit predicted overlays need no changes. Verified live: all overlays (nhi-governance, toxic-findings, attack-path-fusion, cost, aspm, runtime, repo, a2a/mcp) run untouched **except two** that mutated a *held* node subset across passes — cnapp exposure marks and effective-permissions admin-equivalence marks. That hold-then-mutate-later pattern the store's hand-out/write-back-on-eviction contract cannot serve once held objects are evicted. Both were made store-safe with a **minimal, byte-identical** change: single-pass filtering (no full-node materialization) + re-resolving the node through the graph immediately before the mutation (in-RAM this returns the same object, so the default path is unchanged).

## Proof

- **Byte-identical** — builder-into-store vs in-RAM `to_dict()`; the persisted snapshot (nodes/edges/attack_paths/nhi_governance_findings/analysis_status) flag-on vs -off on **both SQLite and live Postgres** persist targets; and both overlays byte-identical under forced LRU eviction (`capacity=4 << N`). The eviction guard was confirmed to bite (RED without the fix, GREEN with it).
- **Producer peak** — the store-backed build removes the whole-node dict + adjacency/reverse-adjacency + dedup indexes. Measured store/in-RAM peak: `≈0.72 @3.6k nodes → 0.52 @7.2k → 0.44 @14.4k → 0.39 @28.8k` (advantage **widens** with scale; in-RAM ≈2.45 KB/node flat vs store falling 1.72→0.98 KB/node).
- **Default-off** — no container ⇒ in-RAM `UnifiedGraph`, identical output.

Guards: `tests/graph/test_store_backed_build_wiring.py`, `tests/api/test_store_backed_build_persist.py`.

## Honest boundary — Part of #4075, not Closes

Byte-identical is fully proven. The producer peak is a **large, measured reduction (~2.7x, widening with scale)** but **not a strict sub-linear bound**: a residual O(N) term survives store-backed — Phase-A's in-pass id-bookkeeping maps plus a few overlays' bounded node subsets stay resident — so peak is a smaller-constant O(N). Fully decoupling peak from N needs streaming Phase-A / two-pass overlays that never hold an O(N) working set. Default stays off until an ops decision flips it.

## Surfaces touched

Build flag + per-build SQLite workspace only. **No** API/CLI/MCP/export/UI/DB-schema surface change. Flag added to the env-var allowlist; ADR-006 records the realized reduction + how to enable.

## Verification

- `ruff check` + `ruff format --check` clean on changed files; `mypy` clean on the 4 changed source files.
- Regression: 59 passed across wiring/persist(sqlite+live PG)/overlays/store-backed/consumer/scale/memory/control-plane. (The 4 failing `test_store_backed_unified_graph_postgres.py` cases are a pre-existing env limitation — the app DB role lacks DDL on schema `public` for the *shared-PG* workspace path this design deliberately avoids; they fail identically on the base commit.)
- Builder+overlay sweep: 101 passed.

Part of #4075.


msaad00:
Superseded by #4314 (combined per the minimal-PR request; branch retained on origin).

PR #4456, opened 2026-07-24
fix(sarif): carry advisory description and remediation into exported rules

msaad00:
## Summary

- Every exported SARIF rule carried a placeholder `fullDescription` of the form `Vulnerability CVE-2020-14343` and no `help` object, so every GitHub code-scanning alert agent-bom produces has a useless body and an empty detail pane. Present in released 0.97.5 — users have it now.
- Root cause: `output/sarif.py` routed advisory prose through `redact_for_persistence(..., EvidenceTier.SAFE_TO_STORE)`. That allowlist exists to protect the runtime-capture path and drops every field name it does not recognise; `description` and `recommendation` are deliberately absent from it, so the text was discarded unconditionally and the fallback always fired. The data was fine — the JSON report carries it verbatim; it just never reached the SARIF.
- Published advisory prose now uses a dedicated export sanitizer (`src/agent_bom/output/advisory_text.py`) instead of the persistence redactor, and rules gain a SARIF 2.1.0 `help` object carrying the description, the remediation, and the advisory reference.

## Verification

Demo scan, `AGENT_BOM_CONFIG=/nonexistent uv run agent-bom scan --demo --offline -f sarif`:

| | before | after |
|---|---|---|
| rules | 18 | 18 |
| placeholder `fullDescription` | **15** | **0** |
| rules with a `help` body | **0** | **18** |
| errors vs vendored SARIF 2.1.0 schema | 0 | 0 |

The new regression suite fails on the pre-fix tree — 15 of 23 red, including
`test_cve_rule_full_description_is_the_advisory_summary`
(`assert 'Vulnerability CVE-2020-14343' == 'PyYAML arbitrary code execution via yaml.full_load on untrusted input'`)
and `test_every_rule_carries_a_help_object`.

Commands run:

- `pytest tests/ -k "sarif"` — 152 passed, 1 skipped
- `pytest tests/ -k "redact or evidence_policy or persistence"` — 199 passed, 5 skipped
- `pytest` on all 25 test files that exercise `to_sarif` — 626 passed
- `ruff check` + `ruff format` on the touched files — clean
- `mypy src/agent_bom/output/sarif.py src/agent_bom/output/advisory_text.py src/agent_bom/evidence/policy.py` — no issues
- `make preflight` — passed
- `git diff --check` — clean

**Runtime evidence redaction is unchanged.** `src/agent_bom/evidence/policy.py` is not touched by this PR — the allowlist has the same 202 entries and `description` / `recommendation` are still off it. Three tests pin that: `test_runtime_evidence_redaction_still_drops_advisory_field_names`, `test_proxy_audit_chain_still_drops_runtime_payloads` (an end-to-end persistence path, not just the helper), and `test_sarif_property_bags_still_use_the_persistence_allowlist`. The pre-existing `tests/test_evidence_policy.py` suite passes unmodified.

## Notes

**Sanitization approach.** The split is by *provenance*, not by output format:

- **Published advisory prose** — OSV/NVD summaries, CIS benchmark recommendations, IaC rule messages, AI-inventory descriptions — is upstream published data, not captured runtime evidence, so it never belonged in the persistence redactor. It goes through `sanitize_advisory_text()`, which keeps the existing `sanitize_sensitive_payload` defences and adds path scrubbing, Markdown-link defanging, and a hard length cap.
- **Scanner free text that can quote the scanned repository** — SAST/secret/credential-exposure descriptions on the unified finding path, and model-authored triage rationale — keeps the conservative allowlist via `_sanitize_scanner_text()` and still falls back to the structural title. This preserves the existing tested guarantee in `test_pr_gate_sarif_action_contract.py` that copied workspace content never reaches an uploaded SARIF; both of those tests pass unmodified.
- **Property bags** (`_sanitize_sarif_property`) are unchanged and keep the allowlist, because they carry arbitrary nested evidence.

**What a hostile advisory string can and cannot do after this change.** It *cannot*: carry a credential (an AWS key, GitHub PAT, JWT, or high-entropy secret collapses the whole string to `***REDACTED***`); carry credentials or a query string inside a URL (stripped to scheme + host + path); disclose the scanning host's directory layout (absolute paths are reduced to `<path:basename>`, and a credential-shaped basename collapses further to `<path:path>`); inject control characters or ANSI escapes into a terminal (stripped); leak an e-mail address (masked to `a***@e***.com`); exceed 1000 characters per field or 4000 for the composed `help`; render a disguised Markdown link in the alert pane (`[label](target)` keeps the label and drops the target, and the one agent-bom-authored `[Reference](…)` link is only emitted for a plain `http(s)` URI with no Markdown-significant characters, so an upstream `remediation.docs` value carrying `)` cannot break out of it).

It *can*: place arbitrary printable prose into the alert body, including a visible bare URL that GitHub will autolink. GitHub renders `help.markdown` through its own sanitizer, so the worst case is misleading text or a link to an attacker-controlled domain the reader can see — a social-engineering surface, not code execution or exfiltration. This is the same exposure `shortDescription` already has today, and the trade is deliberate: the alternative is the shipped behaviour where the field is empty.

**Deliberate trade-off.** Absolute-path scrubbing means a remediation line mentioning `/etc/docker/daemon.json` renders as `<path:daemon.json>`. The basename keeps it intelligible, the verbatim command remains machine-readable in `properties.remediation.fix_cli`, and `helpUri` still points at the authoritative doc.

**Blast radius.** `src/agent_bom/output/skills_sarif.py` has the same helper shape, but its `detail` field embeds workspace snippets by construction (`skill_audit_behavior.py:623` formats `Detected in {filename}: "{snippet}"`), so the allowlist is correct there and it is left alone. Its unrelated raw `properties.context` passthrough is a separate issue and not bundled here.

**Spec.** `reportingDescriptor.help` is a `multiformatMessageString` (`text` required, `markdown` optional) per OASIS SARIF 2.1.0 §3.49.13, confirmed against the spec text and the vendored `tests/fixtures/sarif-schema-2.1.0.json`; the emitted document validates with zero errors.

### DIFF
--- a/src/agent_bom/output/advisory_text.py
+++ b/src/agent_bom/output/advisory_text.py
@@ -1,0 +1,100 @@
+"""Export-side sanitization for published advisory prose.
+
+Advisory text — OSV/NVD vulnerability summaries, CIS benchmark remediation
+guidance, IaC rule messages — is upstream *published* data. It is not captured
+runtime evidence, so it must not be routed through
+:func:`agent_bom.evidence.redact_for_persistence`: that redactor's tier-A
+allowlist protects the runtime-capture path by dropping every field name it
+does not recognise, and ``description`` / ``recommendation`` are deliberately
+absent from it. Sending advisory prose through it discarded the text wholesale.
+
+Export still needs defensive redaction, so this module applies it directly:
+credential-shaped substrings, credentials and query strings inside URLs,
+absolute filesystem paths, e-mail addresses, control characters, and disguised
+Markdown links are removed, and the result is length-capped.
+
+The split is by provenance, not by output format. Scanner free text that can
+quote the repository being scanned — SAST and secret finding descriptions,
+model-authored triage rationale — is *not* advisory prose and keeps the
+conservative allowlist at its own call sites.
+"""
+
+from __future__ import annotations
+
+import re
+from typing import Any
+
+from agent_bom.security import sanitize_path_label, sanitize_sensitive_payload
+
+ADVISORY_TEXT_MAX_LEN = 1000
+ADVISORY_HELP_MAX_LEN = 4000
+
+# Absolute filesystem paths embedded mid-sentence are scanning-host detail, not
+# published advisory content. The lookbehind keeps URL paths (preceded by ``:``
+# or ``/``) and relative traversal such as ``../../etc/passwd`` intact, and the
+# ``{2,}`` floor keeps single generic roots like ``/tmp`` readable.
+_EMBEDDED_ABSOLUTE_PATH_RE = re.compile(r"(?<![\w:/~.\-])(?:(?:/[\w.\-+@]+){2,}/?|[A-Za-z]:[\\/][\w.\-+@]+(?:[\\/][\w.\-+@]+)*)")
+
+# A Markdown link lets an advisory render arbitrary text over an attacker-chosen
+# target in GitHub's alert pane. Keep the label, drop the target; a bare URL is
+# left alone because the reader can see where it points.
+_MARKDOWN_LINK_RE = re.compile(r"!?\[([^\]]*)\]\([^)]*\)")
+
+# Only an http(s) URI with no Markdown-significant characters is safe to inline
+# as a link target; anything else would let the target break out of ``(...)``.
+_SAFE_LINK_URI_RE = re.compile(r"https?://[^\s()\[\]<>\"'\\]+\Z")
+
+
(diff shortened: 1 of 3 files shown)


agent-bom (approved):
Verified the refreshed branch passes the SARIF regression suite, lint, preflight, and now inherits the dependency security fix.

agent-bom (approved):

andres-linero (approved):

PR #4460, opened 2026-07-24
fix(cli): normalize scan push URL and give the gateway a runnable first command

msaad00:
## Summary

- expose `scan --push-url` in core help and normalize control-plane base URLs to `/v1/results/push`
- preserve explicit `/v1/...` endpoints, including reverse-proxy path prefixes and fleet sync
- replace the nonexistent gateway fixture command with the control-plane discovery path
- add a public-command guard for required path arguments that the repository neither ships nor generates

## Verification

- `uv run --extra dev-all pytest -q tests/test_cli_scan.py tests/test_push.py tests/test_public_docs_runnable_commands.py` (166 passed)
- `uv run --extra dev-all ruff check src tests/test_cli_scan.py tests/test_push.py tests/test_public_docs_runnable_commands.py`
- `uv run --extra dev-all make preflight`
- `uv run agent-bom scan --demo --offline --no-auto-update-db` (completed scan and produced the expected finding-driven nonzero exit)


agent-bom (approved):
Verified CLI regression, public-command contract, lint, preflight, and demo-scan evidence is recorded in the PR.

andres-linero (approved):

PR #4462, opened 2026-07-24
fix(tests): restore full-suite lint compliance

msaad00:
Summary
- wrap the two AWS Organizations fan-out fixtures that exceed the enforced 140-column Ruff limit
- keep the test behavior unchanged while restoring the full release lint gate

Verification
- `uv run ruff check src tests`
- `uv run pytest -q tests/test_aws_org_scan_fanout.py` (22 passed)
- `git diff --check`

Notes
- found by the fresh 0.98.0 release audit; no production code changes


agent-bom (approved):
Verified full Ruff, the 22 AWS Organizations fan-out tests, and diff hygiene. The change is formatting-only.

andres-linero (approved):

PR #4482, opened 2026-07-25
fix(cloud): gate the routes that spend the platform's own cloud identity

msaad00:
## The exposure

`/v1/cloud/{provider}/cis-benchmark` does **not** use a tenant's stored connection. It evaluates with whatever ambient credentials the control-plane process itself holds — instance role, assumed role, mounted profile. As shipped it was:

- **on by default**, with no operator opt-in (unlike the inventory surface, which is env-gated per provider)
- reachable at `require_authenticated_permission("scan")` — **analyst in any tenant**
- accepting a caller-supplied `profile`, letting any caller name which host AWS profile to spend

While writing the guard, the test run demonstrated the problem by accident: the request reached the **live AWS SDK** before anything stopped it (`CIS check 2.1.1 failed: AccountId must only contain...` in the log). That output is gone now.

## The part the audit missed

Gating the REST route alone would have been theatre. The MCP `cis_benchmark` tool calls the **same benchmark**, with the same ambient credentials and the same caller-supplied `profile`, and had no gate at all — so the tool was a way around the route guard.

The check therefore lives in `agent_bom.cloud.ambient_credentials`, and both surfaces use it. That is also what keeps REST/MCP shape parity intact, which the existing parity test enforces.

## What changed

| | Before | After |
|---|---|---|
| Default | on | off — `AGENT_BOM_CLOUD_CIS_BENCHMARK=1` to enable |
| Permission | `scan` (analyst, any tenant) | `cloud_ambient` (admin only), a new dedicated permission |
| Profile | caller-supplied | `AGENT_BOM_AWS_PROFILE`, server-side; a request profile now returns 400 |
| MCP tool | ungated | same guard, same envelope |

The disabled envelope carries both `error` (so headless callers degrade exactly as they do for a missing SDK) and `status: "disabled"` (so an operator can tell a disabled surface from a broken one).

**Inventory is deliberately untouched** — it already has per-provider env gates and returns non-secret counts, so it keeps the scan-level permission.

## Two mistakes I made and corrected

1. **I put the gate before input validation**, so a malformed region returned "disabled" instead of 400. Caught by `test_cloud_cis_bad_region_400`. Validation now runs first.
2. **I fixed only REST first**, which broke REST/MCP parity and left the bypass above. Caught by `test_rest_cis_shape_matches_mcp_tool`. That test earned its keep.

I also **changed two existing tests**: `test_cloud_cis_allows_admin_role` and `test_cloud_cis_snowflake_is_wired_not_404` asserted the old always-on default. They now set the flag, so they still test what they were written to test — that the route is wired and degrades without a 404 or 500 — rather than encoding a default this PR intentionally changes.

## Verified

- `tests/test_cloud_ambient_credential_guard.py` — 6 tests, written first, confirmed red; includes one asserting the provider is **never invoked** while disabled
- `tests/cloud/` + cloud surface parity + guard — **734 passed**
- `pytest -k "mcp_tool or compliance or rbac or auth"` — **1370 passed**
- All **17** CI gates; `mypy` clean on 4 modules; `ruff` clean

Note on the suite: a broad `-k` sweep shows ~19-23 failures on **unmodified main** too, in different files per run — the repo's known order-dependent test class. I verified against a clean `git stash -u` baseline and re-ran the affected suites with `-p no:randomly`, where everything passes.

### DIFF
--- a/scripts/env_var_allowlist.txt
+++ b/scripts/env_var_allowlist.txt
@@ -60,6 +60,8 @@ AGENT_BOM_AWS_EVENT_QUEUE_URL
 # Opt-in (default OFF) gate for the estate-wide read-only Azure asset inventory scanner (token/credential auth only).
 AGENT_BOM_AWS_INVENTORY
 # Defensive cap on the number of member accounts a single AWS Organizations scan fans out over (default 200).
+# Operator-selected AWS profile for ambient-credential evaluation; read live and never caller-supplied, so a request cannot pick which host credential the control plane spends.
+AGENT_BOM_AWS_PROFILE
 AGENT_BOM_AWS_MAX_ACCOUNTS
 # Defensive cap on the number of regions a single multi-region AWS inventory scan fans out over.
 AGENT_BOM_AWS_MAX_REGIONS
@@ -94,6 +96,8 @@ AGENT_BOM_CLICKHOUSE_MAX_QUEUE
 AGENT_BOM_CLICKHOUSE_ACCESS_TOKEN
 AGENT_BOM_CLICKHOUSE_USER
 # Opt-in (default OFF) gate for the estate-wide read-only AWS asset inventory scanner.
+# Opt-in (default OFF, read live) gate allowing this control plane to run the CIS benchmark against its OWN ambient cloud identity rather than a tenant connection.
+AGENT_BOM_CLOUD_CIS_BENCHMARK
 AGENT_BOM_CLOUD_INVENTORY
 # Opt-in (default OFF) bounded S3 object byte-range sampling for DSPM content classification. Stores redacted classification counts, never raw object bytes.
 AGENT_BOM_DSPM_S3_SAMPLING
--- a/src/agent_bom/api/routes/cloud.py
+++ b/src/agent_bom/api/routes/cloud.py
@@ -37,6 +37,12 @@ from fastapi import APIRouter, HTTPException, Query, Request
 from agent_bom.api.models import RuntimeEvidenceIngestRequest
 from agent_bom.api.tenancy import require_request_tenant_id
 from agent_bom.backpressure import BackpressureRejectedError, adaptive_backpressure
+from agent_bom.cloud.ambient_credentials import (
+    PROFILE_REJECTED_NOTE,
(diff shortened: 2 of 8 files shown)


msaad00:
Pushed `fc4e48bf2`, which fixes the two failures that belong to this PR: `test_cloud_cis_benchmark_offloads_evaluation` and `test_cloud_cis_timeout_returns_bounded_unavailable_envelope` asserted the old always-on default, so they now enable `AGENT_BOM_CLOUD_CIS_BENCHMARK` and go on testing the offload and timeout envelope rather than the guard.

The other three failures in that run — `test_readme_storefront_is_concise_ordered_and_actionable`, `test_public_docs_do_not_overclaim_smithery_catalog_liveness`, `test_release_prep_does_not_call_unpublished_version_current` — are **not from this branch**. They came in with #4480 and are red on `main` itself. #4483 fixes them. This PR will go green once that merges and this branch is rebased.

msaad00:
Ran CI's exact Test (Python 3.13) command locally against branch tip `b61f3e24d`, in a CI-faithful environment (`uv sync --frozen --extra dev --extra api --extra mcp-server`):

```
16495 passed, 177 skipped, 0 failed in 256.87s
```

**Correction to my earlier note on this PR.** I attributed a batch of local failures to "the repo's known order-dependent test class." That was only partly right. A first run here reproduced 4 failures *in isolation* — so not order-dependent — but they turned out to be **local-environment artifacts**: the worktree venv lacked pytest, so `uv run pytest` fell through PATH to a stale venv carrying partial extras (`databricks-sdk` without `databricks-sql-connector`, `opentelemetry-api/sdk` without the OTLP exporter). Each of those tests is gated by `pytest.importorskip`, and CI installs neither extra, so all four skip there. In the CI-faithful env the suite is clean.

The lesson: a local failure is not evidence of a real failure unless the environment matches CI's lockfile. Worth knowing before anyone chases those four.

This branch's own tests — `test_cloud_ambient_credential_guard.py`, `test_cloud_scan_offload.py`, `test_api_cloud_surface_parity.py` — 43 passed.

andres-linero (approved):

agent-bom (approved):

PR #4486, opened 2026-07-26
perf(jobs): cache the status counts the dashboard polls

msaad00:
## The cost

`count_summary_by_status` is an unbounded `GROUP BY` over `scan_jobs` — **measured at ~342ms across 200k rows** — and the activity feed asks for it every few seconds, for every open dashboard. The search variant is worse: its predicate lowercases a `CONCAT_WS` over several columns plus a jsonb path, so it can never use an index and the cost tracks table size no matter how narrow the result.

Combined with the `list_jobs` offload from #4476, that was 342ms of database work per poll per viewer.

## The fix

A five-second in-process TTL cache, mirroring `findings_count_cache`, which already solves the same problem for finding totals.

Three properties that matter:

- **A job write invalidates immediately.** `put()` and `delete()` drop the tenant's entries, so a finished job is never hidden behind a stale count. Freshness of *state* is preserved; only repeated identical polls are absorbed.
- **The search term is part of the key.** A filtered count is never served as the unfiltered one, or vice versa — asserted both at the cache level and through the route.
- **Reads return a copy.** A caller editing the response cannot corrupt the cached entry.

`AGENT_BOM_JOB_STATUS_COUNT_CACHE_TTL_SECONDS` tunes it; `0` disables caching entirely for deployments that want every poll live.

## Verified

- `tests/test_job_status_count_cache.py` — 10 tests, written first. Includes two route-level integration tests that assert the aggregate runs **once** across two polls, and **twice** when the search term differs — i.e. they prove the cache is actually consulted, not just that it exists.
- `tests/test_hardening.py` (job pagination) + cache — 31 passed
- All 17 CI gates; `mypy` clean on the three changed modules; `ruff` clean
- New env var declared in the allowlist, env reference regenerated

Last item from the audit's scale lane. The remaining known read-path cost is the findings `COUNT(*)` on TTL expiry, which is a different shape (single-flight rather than caching) and is left for its own change.


msaad00:
Pushed `ebea8bace`. A pre-review CI-parity run caught **two real problems in this PR**, both now fixed.

**1. The cache leaked across tests.** `_entries` is process-global, so a count captured against one test's store was served to the next test on the same worker — which swaps the store but not the cache. `test_list_jobs_response_contract_is_unchanged` was seeing another file's `{'done': 4}`. It reproduced even running that one file alone. `tests/conftest.py` already resets process-global state for exactly this class of bug, so both read-path count caches now reset there.

**2. The module overclaimed.** It said a finished job "never sits behind a stale count." That holds only within one process. Across a multi-worker control plane, a write handled by worker A leaves worker B serving its own cached counts until the TTL expires — so the honest guarantee is **bounded staleness**, not immediate consistency. That is fine for a polled progress display, which is why the window is seconds, but the docstring now states it plainly instead of implying more.

The second one is the one worth reviewing carefully: the behaviour is unchanged and acceptable, but the original wording would have left an operator believing something stronger than the code delivers.

agent-bom (approved):

andres-linero (approved):

PR #4490, opened 2026-07-26
fix(auth): enforce roles in no-auth mode

msaad00:
## Summary

- keep the authorization middleware active when anonymous access is explicitly enabled, so credential-less callers traverse the configured role matrix
- enforce Viewer, Analyst, and Admin behavior for source, schedule, and scan routes, with the hosted demo always clamped to Viewer
- reject spoofed or unattested proxy identity headers instead of silently downgrading them to anonymous access

## Verification

- `uv run --no-sync pytest -q tests/api/test_no_auth_role_enforcement.py` — 13 passed
- focused auth, source, schedule, scan, and blueprint suites — 274 passed
- broader auth/role/health selection — 106 passed
- `make preflight`
- `make lint`
- `git diff --check`

## Notes

Pure no-auth operation remains an explicit opt-in. Authentication may be disabled for a self-hosted deployment, but authorization is still enforced through `AGENT_BOM_NO_AUTH_ROLE`; the default effective role is Viewer.


msaad00:
Pushed `4601e1804` fixing `tests/test_exports_api.py` (11 passing). **But the blast radius is much wider than the 8 failures CI surfaced, and that is the important part of this comment.**

## What the change actually means

Keeping `APIKeyMiddleware` installed in pure no-auth mode redefines an unauthenticated call: it now resolves to `NO_AUTH_ROLE` and traverses the route role matrix, instead of being rejected because the middleware was absent. That is a *better* posture — authorization applies where it previously did not — but it invalidates the assumption `unauthenticated ⇒ 401` everywhere it is asserted.

Measured, main vs this branch, same command (`-p no:randomly`):

| Suite | main | branch |
|---|---|---|
| `tests/test_ticketing_route.py` | 9 passed | **8 failed** |
| `tests/test_exports_api.py` | 10 passed | 10 failed → **11 passed** after this commit |
| `-k "auth or export or api_"` sweep | — | **46 failed**, 1933 passed |

The failures are all one of two shapes: `assert 200 == 401` (unauthenticated read now allowed as NO_AUTH_ROLE) or `assert 401 == 403` (rejected by role rather than by missing credentials).

## Second, separable bug this exposed

The export fixture set its own auth environment but never rebuilt the middleware stack, so the app kept whatever posture the previous test on that worker left behind — which is why trusted-proxy headers were rejected. Reproduced outside pytest returning 200, inside pytest returning 401. Fixed by calling `configure_api_from_env()` at fixture setup and teardown. **That pattern is what the other suites need too**, not a per-assertion edit.

## Correction to something I said earlier in this thread

I first called this a regression. It is not — it is the intended contract change, and the security posture is stronger. I was wrong and want that on the record before anyone acts on it.

## What I would decide before merging

The 401→200 change is user-visible for anyone running no-auth mode who currently relies on "no credentials means no access". If that is intended, the remaining suites need the same fixture treatment and the behaviour should be called out in the changelog. Happy to do that sweep — say the word.

agent-bom (dismissed):

andres-linero (dismissed):

agent-bom (approved):

andres-linero (approved):

PR #4535, opened 2026-07-28
fix(security): enforce trusted maintenance and trial boundaries

msaad00:
## Summary

- require distinct application and marker-authorized maintenance Postgres principals for cross-tenant work, with a bounded maintenance pool and no app-pool fallback
- force tenant RLS on the dispatch queue, move global callers to explicit maintenance connections, and preserve queued rows through forward-only migration 20260728_03
- authorize managed-trial operators before generating invitation material and offload blocking invitation, lifecycle, and session reads from the event loop
- wire separate application, maintenance, migration/admin, and auth secrets through Compose and staged Helm production profiles

## Verification

- `uv run pytest -q` — 16,774 passed, 185 skipped
- focused security/deploy regression matrix — 419 passed, 3 skipped
- live Postgres integration — 16 passed; graph workspace — 5 passed
- exact legacy pre-0.98.3 database upgrade preserved queue rows and enforced scoped maintenance RLS
- `uv run ruff check src tests deploy/supabase/postgres/alembic/versions/20260728_03_trusted_maintenance_rls.py`
- `uv run mypy src/agent_bom` — 813 source files clean
- `make preflight`
- actionlint on modified workflows
- all three Compose configurations validated
- production secret-sync and workload Helm profiles linted and rendered
- live backup/restore smoke used a NOSUPERUSER NOBYPASSRLS maintenance login with FORCE RLS

## Notes

- SQLite deployments are unchanged.
- The security migration is intentionally forward-only; downgrade refuses instead of restoring the vulnerable bypass.
- The explicit superuser acknowledgement remains limited to disposable single-tenant development and never permits app-pool maintenance access.
- AWS platform credential bootstrap and Terraform wiring remain scoped to the later deploy PR in the approved series.


agent-bom (approved):
Independent exact-head review completed. Maintenance identity separation, scoped RLS bypass, queue preservation, trial authorization/offload, deployment secret wiring, migration behavior, and the CI follow-up were reviewed with no remaining P0/P1/P2 findings. Focused validation: 148 tests passed; Ruff, mypy, actionlint, and diff checks passed. Credentialed Postgres and full-suite checks are running on the exact head.

agent-bom (approved):

andres-linero (approved):

PR #4575, opened 2026-07-30
perf(control-plane): bound, de-duplicate, and honestly report the findings read path

msaad00:
## Summary

Read-path hardening from a hands-on audit. Five defects that each made the control plane
either report a wrong number or fall over under ordinary analyst load.

- **Bounded scope walks now say they were bounded.** PR #4566 added the scan budget and
  deadline to `collect_scope_filtered_page`, but all three `list_current_page` callers
  discarded the truncation metadata, so a sparse filter returned an empty page with a
  "more" cursor and no partial signal at all. The metadata is now plumbed through the
  Protocol and every backend, surfaced as `scope_completeness` plus an explicit warning,
  and accumulated across refills rather than overwritten by the last one.
- **`/v1/overview` no longer stampedes.** The cache was a plain read-compute-write with no
  in-flight de-duplication; 8 concurrent readers each computed the full fold. Now
  single-flight per tenant. Measured 8 folds -> 1.
- **Read concurrency is no longer pinned at 8.** The `findings` controller's ceiling was a
  literal 8 regardless of hardware, which shed ~92% of reads at 50 clients. It now derives
  from `WORKER_THREAD_LIMIT` (default 20), matching the worker pool the reads actually run
  on. Env override and shedding behaviour unchanged.
- **Bulk ingest counts distinct rows.** `new_rows` counted each row against the DB but never
  deduped within the batch, while `ON CONFLICT DO UPDATE` collapsed intra-batch duplicates
  and the cached tenant total advanced by the inflated count. Posting 3 rows sharing a
  canonical id reported 3 ingested while 1 landed.
- **The retention tick runs off the event loop.** Six of eight work items in `_cleanup_loop`
  ran inline, including Postgres partition DDL, which would freeze the control plane every
  60 seconds at volume.

The scan-budget/deadline bounds are now resolved per call so operators can tune them
without a restart.

## Full-stack alignment

`scope_completeness` is a new response field, threaded through `ui/lib/api-types.ts` and
rendered in `ui/app/findings/page.tsx` as a partial-results notice. The findings page also
stops showing "No findings found" when the walk was partial -- that empty state was the
user-visible face of this bug.

## Verification

- targeted `pytest`: 90 passed (read-scale, facets export contract, scope-filter keyset,
  envelope contract, backpressure, overview, overview cache, new hardening module);
  hub/server suites 182 passed, 1 skipped
- every new guard verified red before green
- `ruff check` clean on all 9 changed Python files
- `mypy src/`: 12 errors, all pre-existing cloud-SDK stub mismatches in files not in this diff
- preflight drift gates: OpenAPI OK, v1 schemas OK, product surface passed, env-var reference OK
- UI: lint clean, typecheck clean, `vitest` 978 passed across 163 files

## Notes

- **Postgres changes are verified by symmetry only.** The local PG instance was unavailable.
  Both PG edits mirror the SQLite path exactly and the PG scope path is exercised by the
  existing fake-connection test against the real store class. The distinct-id count assumes
  `executemany` + `ON CONFLICT DO UPDATE` collapses intra-batch duplicates on PG as on
  SQLite; that is the one assumption not reproduced empirically.
- Raising the concurrency default lets more reads through, so the 12s p99 cooldown now
  carries more of the shedding load. Intended, but worth watching under real load.
- `ruff format --check` flags 3 files, all in lines this branch never touched; `make lint`
  runs `ruff check` + `mypy` only, and main is not format-clean either. Left alone rather
  than sweeping unrelated reformatting in.

### DIFF
--- a/scripts/env_var_allowlist.txt
+++ b/scripts/env_var_allowlist.txt
@@ -330,6 +330,12 @@ AGENT_BOM_OCI_MAX_LAYER_UNCOMPRESSED_BYTES
 AGENT_BOM_OFFLINE
 # TTL (seconds) for the per-tenant /v1/overview payload cache; read at call time so tests can vary it. 0 disables.
 AGENT_BOM_OVERVIEW_CACHE_TTL_SECONDS
+# How long a concurrent /v1/overview reader waits on the in-flight fold before recomputing itself; read at call time.
+AGENT_BOM_OVERVIEW_SINGLEFLIGHT_TIMEOUT_SECONDS
+# Rows one scope-filtered /v1/findings page walk may examine before it stops and reports a partial page; read at call time.
+AGENT_BOM_SCOPE_FILTER_SCAN_BUDGET
+# Wall-clock ceiling (seconds) for one scope-filtered /v1/findings page walk; read at call time.
+AGENT_BOM_SCOPE_FILTER_DEADLINE_SECONDS
 # TTL (seconds) for the hub-derived severity cache; read at call time so tests and operators can disable it with 0.
 AGENT_BOM_HUB_OVERVIEW_CACHE_TTL_SECONDS
 AGENT_BOM_SKIP_UPDATE_CHECK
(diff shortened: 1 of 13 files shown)


msaad00:

PR #4577, opened 2026-07-30
fix(graph): stop reporting bounded and sampled results as complete

msaad00:
## Summary

Four defects where the graph reported a number its own evidence did not support. The theme
is the same each time: a bounded or sampled result presented as complete fact.

- **Roll-ups ignored source truncation.** `df0fc5997` added `_inherit_completeness` to
  `filtered_view` only, so `rollup_view`, `drill_down` and `attack_path_view` computed
  completeness from their own orphan-limit cut and discarded the snapshot's. A 61-node
  estate budgeted to 15 reported `{"status":"complete","complete":true}` beside
  `summary.total_nodes: 15`. This was latent because `/v1/graph/rollup` passes no node
  budget today -- which is also why it is slow at scale, so the obvious perf fix would have
  silently switched the lie on.
- **Five views dropped the flag entirely.** `inventory_view`, `attack_path_view`,
  `lateral_movement_view`, `compliance_view` and `runtime_view` all returned `_subgraph(...)`
  directly, reporting `truncated: false, total: 0` over a truncated source.
- **Sampled bottleneck rankings were presented as fact.** The estimate sampled the first 50
  nodes in insertion order. On a 120-agents -> one CHOKE server -> 120-packages estate with
  vulnerability nodes inserted first -- exactly what a real build produces -- it returned
  five score-`0.0` CVE nodes and omitted the actual choke point. `total = sum(scores) or 1.0`
  also normalized an all-zero result into zeros that rendered as legitimate scores, and
  `degree_centrality` counted only out-edges, so a node with 50,000 in-edges scored 0.
- **`stats` contradicted `completeness` in the same response.** The docstring claimed *"Stats
  reflect the full (unpaginated) graph"*; the payload carried `completeness.total: 100000`
  beside `stats.total_nodes: 10826`.

## Verification

- 70 graph suites + new tests + `test_cli_scan.py`: **982 passed, 11 skipped** (11 skips are
  Postgres-gated; the local instance was down)
- `tests/graph/` 38 passed, 9 skipped · `tests/test_graph_api.py` 66 passed
- `ruff check` and `ruff format --check` clean on every changed file
- `mypy` clean on all 7 changed modules
- `scripts/export_openapi.py --check` flagged drift; regenerated and confirmed by diff that
  the only change is the corrected `/v1/graph` description. Artifacts committed.
- UI: `npm run typecheck` clean, `vitest` 163 files / 976 tests passed

## Semantic changes -- please read before merging

- `degree_centrality` now counts distinct neighbours in **either** direction rather than
  out-edges. Bidirectional edges no longer double-count and values are bounded by 1.0.
- `bottleneck_nodes` **omits score-0.0 entries**, so results are often shorter than `top_n`;
  an all-zero graph returns `[]` where it previously returned a full ranking of zeros.
- `completeness.reason` may now be comma-separated when two cuts applied
  (`"node_budget,orphan_limit"`); the UI renders them as `A · B`.
- `_inherit_completeness` falls back to the source node count when `total_nodes` was never
  populated, so `total: 0` can no longer sit beside a non-zero `returned`.
- `GraphBackend` gained `bottleneck_analysis`; `bottleneck_nodes` keeps its existing
  signature, so no caller changes are required.

## Notes

- The bottleneck bug existed byte-identically in **three** implementations. The CLI path
  (`scan_cmd.py:2159`) calls `GraphBackend.bottleneck_nodes`, not the `UnifiedGraph` method,
  so fixing only the latter would have left the shipped CLI output wrong. All three now sit
  behind one shared algorithm, and `NetworkXBackend` declares `sampled_sources ==
  total_nodes` rather than leaving callers to infer which backend answered.
- The view-inheritance test parametrizes off `*_view` methods discovered on the class, and a
  companion assertion fails if a new view is added without being declared -- so a sixth view
  cannot silently skip the flag.
- **Regression found and fixed inside this branch:** `StoreBackedUnifiedGraph` bypasses the
  dataclass `__init__` and never mirrored `completeness`, so once `stats()` read it every
  store-backed `to_dict()` raised `AttributeError` (14 failures). Fixed, plus
  `test_store_backed_field_parity.py`, which walks `dataclasses.fields(UnifiedGraph)` so the
  next added field fails there instead of in production.
- `npm run build` was not run (its standalone step exhausted the disk earlier in this
  session). The UI delta is three additive type fields and one string formatter; request the
  build gate before merge if you want it covered.
- Two `ruff format` deviations in `rollup.py` and `api/routes/graph.py` sit outside this diff
  and are left untouched -- local ruff is 0.16.0 while pre-commit pins 0.9.7.


msaad00:

PR #4578, opened 2026-07-30
fix(scan): fail closed when a version bound cannot be compared

msaad00:
## Summary

Six defects in the vulnerability-matching and enrichment layer, from a hands-on audit. The
common shape: when a version could not be compared, the code included the finding anyway.

- **Two range engines with opposite fail policies.** `_version_in_window` carried its own
  copy of the range walk that *skipped* an uncomparable bound (treating the package as
  affected), while `version_in_range` fails closed and says so. Same package, different
  verdict depending on which path served the query. `_version_in_window` now delegates to the
  fail-closed engine; `introduced: "0"` maps to "no lower bound" so open-ended ranges do not
  become false negatives.
- **Maven qualifier versions were uncomparable**, so every `.RELEASE`-bounded advisory fell
  through as a false positive. Implemented against the Apache POM Reference "Version Order
  Specification": tokenization on `.`/`-`/`_` and digit-alpha transitions, trailing-null
  trimming, and the qualifier order `alpha < beta < milestone < rc = cr < snapshot < "" =
  final = ga = release < sp`.
- **The GHSA path ignored the advisory's lower bound** whenever a fixed version was
  extractable, and `_installed_version_is_affected` parsed every ecosystem with PEP 440 and
  returned `True` on any parse error. Both now fail closed with a logged reason, and the
  range is always evaluated with `fixed` used only as a short-circuit.
- **`kev_due_date` was dropped on the local-DB read path.** The data was present and correct
  and every consumer was already wired -- only the reader omitted it, so the BOD 22-01
  remediation deadline was silently unavailable to every downstream format.
- **Reachability claimed function reach from a bare class token.** `_symbol_tokens` expanded
  both sides, so an advisory naming `Environment.from_string` was satisfied by any use of
  `Environment`. Expansion is now asymmetric -- head-token expansion applies only to the
  reached side. The module's own docstring already said *"we cannot prove function
  reachability, so we never claim it."*
- **The posture grade asserted "No vulnerabilities found" when the vuln DB was
  unavailable.** The guard measured discovery, not evaluability. New shared
  `vulnerability_coverage_incomplete()` scoped to vulnerability-lane issue codes, so a failed
  cloud collector does not blank an otherwise sound grade.

### Measured

`agent-bom check org.springframework:spring-core@5.3.20 -e maven`, run against an
`origin/main` baseline side by side:

```
before: 5 findings -- 1 true, 4 false   (CVE-2009-1190, CVE-2018-1258, CVE-2018-15756, CVE-2018-11040)
after:  1 finding  -- CVE-2025-41249     (exactly what live OSV returns for that version)
```

Every false positive was a `.RELEASE`-bounded advisory; CVE-2009-1190 was being reported
against a 2022 release.

## The accuracy gate caught a regression, and the corpus was the problem

The first sweep went red: recall 1.0 -> 0.9934, 72 misses. Root cause was the corpus oracle,
not the matcher. All 72 come from three advisories carrying a `GIT` range at `introduced: 0`
beside an ECOSYSTEM range with a non-zero lower bound -- OSV's importer walks the repo and
writes every reachable tag into `versions[]`, so the "ground truth" named `requests v0.2.0`
while NVD scopes CVE-2015-2296 to *"2.1.0 through 2.5.3"*. The previous recall of 1.0 was
bought with 28 false positives.

`scripts/cve_matching_accuracy.py` now skips self-contradictory blocks as
`skipped_unsound_oracle` and reports the count, with `--check` failing if it grows. **The
predicate reads only the advisory data and never the matcher's answer**, so it cannot be
tuned to flatter the score. 6 blocks of 127 skipped; multi-window coverage stays at 22 and
the mutation test still detects a window collapse. Re-derived baseline: **recall 1.0,
precision 0.9974 -> 1.0, evaluated 127 -> 121.**

## Correction to an earlier claim

An intermediate report asserted this branch took `golang.org/x/net@0.16.0` from 17 findings
to 16, matching live OSV. **That does not reproduce and is withdrawn.** The exhaustive
re-derivation -- old vs new comparator over the real local DB (14,602 Go advisory windows,
1,535 packages, every version each package's own advisories use as a bound) -- gives:

```
OLD comparator: 172808 matches (6906 distinct advisory IDs)
NEW comparator: 182021 matches (7084 distinct advisory IDs)
removed: 858    added: 10071    advisory IDs dropped: 0
```

Net *more* matches, in the opposite direction, with zero advisory IDs lost. Both flip classes
were spot-checked against real windows: the 858 removed are pre-release pseudo-versions below
their window's lower bound that the old code collapsed to the base tag (correct removals),
and the 10,071 added are form-3 pseudo-versions that were previously unrecognized, hence
uncomparable, hence fail-closed misses (recovered true positives). The pinned case still
holds: `x/net 0.16.0` vs `GO-2023-1495` is correctly out of range.

## Verification

- 23-suite targeted sweep: **522 passed, 0 failed**
- `ruff check` clean on all 10 changed files; `mypy` **Success: no issues found in 7 source
  files** (fixed a real `[index]` error introduced at `version_utils.py:607`)
- `pre-commit run --from-ref origin/main --to-ref HEAD`: ruff, bandit, env-drift,
  comment-hygiene, json, eof, whitespace all pass

## Notes

- `ruff-format` and the `mypy` hook fail, **both pre-existing and proven so**: each flagged
  file was extracted at `origin/main` and re-run against the pinned `ruff@0.9.7` -- all three
  reformat on main, and none of the hunks touch lines in this diff. The 39 `no-any-return`
  errors are all in `api/routes/*`, `middleware.py` and `server.py`, none of which this
  branch touches; `pre-commit run mypy --files src/agent_bom/api/routes/cloud.py` alone
  reproduces the identical 39.
- `db/lookup.py:471`'s `unknown -> include` policy is deliberately unchanged -- it is a third
  fail-open engine and deserves its own scoped change.
- Pre-existing and not folded in: `_warn_unparseable_bound` blames the bound when the
  uncomparable operand was the installed version. The verdict is right; only the log line is
  wrong.
- Maven and Go orderings are implemented against their official specifications and pinned by
  spec-verbatim tests, but were not cross-checked against a live `ComparableVersion` or
  `golang.org/x/mod` binary.

### DIFF
--- a/docs/CVE_MATCHING_ACCURACY.json
+++ b/docs/CVE_MATCHING_ACCURACY.json
@@ -1,14 +1,15 @@
 {
   "advisories": 127,
-  "evaluated": 127,
+  "evaluated": 121,
   "skipped_no_ranges": 0,
   "skipped_uncomparable": 0,
+  "skipped_unsound_oracle": 6,
   "multi_window_advisories": 22,
-  "true_positive": 10875,
+  "true_positive": 10622,
   "false_negative": 0,
-  "true_negative": 198,
-  "false_positive": 28,
+  "true_negative": 205,
+  "false_positive": 0,
   "recall": 1.0,
-  "precision": 0.9974,
-  "f1": 0.9987
+  "precision": 1.0,
+  "f1": 1.0
 }
--- a/scripts/cve_matching_accuracy.py
+++ b/scripts/cve_matching_accuracy.py
@@ -50,6 +50,7 @@ class Outcome:
     evaluated: int = 0
     skipped_no_ranges: int = 0
     skipped_uncomparable: int = 0
+    skipped_unsound_oracle: int = 0
     multi_window: int = 0
     true_positive: int = 0
     false_negative: int = 0
@@ -79,6 +80,7 @@ class Outcome:
             "evaluated": self.evaluated,
             "skipped_no_ranges": self.skipped_no_ranges,
             "skipped_uncomparable": self.skipped_uncomparable,
+            "skipped_unsound_oracle": self.skipped_unsound_oracle,
             "multi_window_advisories": self.multi_window,
             "true_positive": self.true_positive,
             "false_negative": self.false_negative,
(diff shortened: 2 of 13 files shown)


msaad00:

PR #4598, opened 2026-08-02
test(ui): stop the graph screenshot helper failing the suite

msaad00:
## What

`captureRenderedRegion` in `ui/e2e/large-graph-overview.spec.ts` called `scrollIntoViewIfNeeded()` **before** the geometry poll. That call moves inside the try/catch it belongs in, next to the screenshot it exists to position, and gets its own timeout. The geometry poll now runs first and stays the real assertion.

## Why

That call waits for the element to hold a stable bounding box — which a continuously animating sigma canvas never does — and it acts on an already-resolved handle that a re-render can detach. CI failed with `element is not stable` followed by `Element is not attached to the DOM`, at exactly that line, while the other 30 specs passed.

Neither condition means the graph failed to render. The file already documents the screenshot as "a debugging artifact, not an assertion — every functional claim about this region is already covered by the canvas-pixel checks", so a failure to position it must not fail the suite. The geometry poll re-resolves the locator on every attempt, so it rides out the re-renders that detach a resolved handle.

This flake has now blocked **two** unrelated PRs (#4592 and #4596), which is why it is worth landing on its own rather than being carried per-branch.

## How verified

- `tsc --noEmit -p ui/tsconfig.json` — exit 0, **0 errors** across the project.
- Failure mechanism is pinned directly by the CI logs on both affected PRs: same spec, same line 318, same two Playwright errors in sequence.
- E2E behaviour itself is verified by CI — the local UI build is not run here (its standalone step exhausts the disk on this machine).


msaad00:
Already on `main`: this fix was merged into #4596 to unblock it, and #4596 has since landed. `git diff origin/main...fix/e2e-graph-capture-stability` is empty and `scrollIntoViewIfNeeded` sits inside the try/catch at `ui/e2e/large-graph-overview.spec.ts:338` on main. Closing as redundant rather than merging an empty diff.

PR #4602, opened 2026-08-02
chore(release): prepare 0.98.3 — version, docs, and re-captured screenshots

msaad00:
Release preparation for **0.98.3**. Intended to merge **last**, after #4601, so the version bump, screenshots and metrics all sit on top of the final tree.

Deliberately a patch bump rather than 0.99.0 — there is no need to approach 1.0 yet.

## Version

Bumped across the **145 managed references** the tooling tracks. `bump-version.py --check` and `check_version_alignment.py` both pass. The OpenAPI documents are regenerated (only `info.version` moved), and the Docker storefront's example-tag row follows; the release workflow renders that row to "current stable" at publish time, so the checked-in copy keeps the lifecycle-neutral wording.

## Paid-tier and hosted-offering copy removed

`EDITIONS.md` described lanes in commercial terms that do not reflect anything shipping: a *"licensed enterprise tier"* on the roadmap, a *"Gated hosted POC"* lane, a managed *"agent-bom Cloud"*, and *"no paid tier in this repo **yet**"* — where the "yet" implies one is coming. The page now states what is true and nothing more: agent-bom is free and open source, and you run it in your own boundary. The cost column and the hosted-POC row are gone from the lanes table, the page is retitled to deployment lanes, and the four inbound links plus the two remaining `agent-bom Cloud` mentions follow.

`docs/AGENT_CAPABILITY.md` is **generated**, so the roadmap-tier line was removed from `scripts/generate_agent_capability_manifest.py` and the manifest regenerated. Editing the rendered file alone had left the generator still emitting it, which is what `generate_agent_capability_manifest.py --check` was reporting as stale.

The enterprise support and procurement docs are deliberately left alone: they exist to *disclaim* paid contracts and hosted SLAs for self-hosted deployments, so they are the opposite of a commercial claim.

## Screenshots re-captured, not re-stamped

The release-consistency gate compares each screenshot's `visible_version` against the version **actually drawn in the image** — the dashboard chrome renders `Control plane · v<version>` on every page. Editing the manifest field would have pushed 21 images that visibly read `0.98.2` past a gate built to catch exactly that.

All 21 were re-captured by the deterministic Playwright harness against the local committed standalone build (which refuses to run on a dirty tree), and the result was checked visually rather than trusted from the manifest.

## CHANGELOG

The staged Unreleased notes describe what actually shipped, so they move under a dated `0.98.3` heading with the compare links rebased, and a fresh Unreleased section opens above.

## Verification

Full preflight — **10/10 PASS**: version alignment, release consistency, package layout, public-docs hygiene, OpenAPI, v1 schemas, env-var reference, capability manifest, check-counts, bump-version.


msaad00:
Already on `main`. The read-path lane had rebased onto this branch rather than `main`, so folding it into #4601 carried all seven release-prep commits along and they merged together.

Verified rather than inferred from the empty diff — on `main` @ `b8dace1c3`:
- `pyproject.toml` → `version = "0.98.3"`
- `docs/images/product-screenshots.json` → `release_version: 0.98.3`
- `CHANGELOG.md` → dated `## [0.98.3]` heading present

Full release preflight on `main` is 10/10: version alignment, release consistency, package layout, public-docs hygiene, OpenAPI, v1 schemas, env-var reference, capability manifest, check-counts, bump-version.

Closing as redundant rather than merging an empty diff.

PR #4647, opened 2026-08-04
feat(output): surface the integrity and provenance verdict everywhere

msaad00:
## Summary

`Package.integrity_verified` and `provenance_attested` were populated by
`--verify-integrity` and then reached **no machine-readable output at all**:

```
provenance verdict dropped by: ['csv', 'cyclonedx', 'json', 'sarif', 'spdx2', 'spdx3']
```

For a supply-chain scanner that is the signal a machine consumer most needs, and
only a human reading the console could get it. The verdict now flows from one
shared derivation into JSON, CycloneDX, SPDX 2.x, SPDX 3.0, SARIF, CSV, finding
evidence and `/v1/findings`.

**Found en route:** the MCP `scan` tool assigned `pkg.integrity` — an attribute
**no model field or emitter reads** — and never checked provenance at all, so
`verify_integrity=True` over MCP produced nothing whatsoever. CLI and MCP now
share one `integrity.verify_packages()` helper, which also closes a latent CLI
gap where the verdict reached only the first `Package` per identity, leaving
duplicates across servers `None`.

## Spec placement

Verified against the versions this repo actually emits, not from memory:

- **CycloneDX 1.7** — `component.evidence.identity[]`, techniques
  `hash-comparison` (integrity) and `attestation` (provenance), confidence 1.0
  pass / 0.0 documented failure. Checked against the repo's own vendored 1.7
  schema by reading its `componentIdentityEvidence` technique enum; output
  validates under `Draft7Validator`.
- **SPDX 3.0.1 / 2.x** — annotations with the spec-designated `other` type.
  `secureSoftwareAttestation` was considered and **rejected**: it is SSDF-scoped
  and expects a locator URL we do not have.

Neither format models a verification *verdict* natively; both use their
designated extension point. Absent means never checked; a failure is emitted
explicitly rather than omitted.

## Two accuracy defects closed alongside

**`/v1/findings` facets could contradict their own total:**

```
AssertionError: severity facet says 100 but total says 10
AssertionError: assert 12 == 3   (facets.severity.critical vs total)
```

The pushed-down histogram is exact and unbounded while `total` came from the
row-budgeted walk. Both now derive from one basis, with per-dimension
`exact`/`bounded` labelling and a warning naming which counts are lower bounds.

**The console mislabelled a package-CVE-scoped count as covering every category:**

```
AssertionError: package-CVE-scoped row must say so, got 'Critical findings'
```

A prior audit called this a non-finding by reading the compact headline — a
mutually exclusive branch. Under `--verbose` both render seven lines apart:
`Critical findings 2` (blast-radius scoped) and `SECURITY POSTURE: 3 CRITICAL`
(all categories). Relabelled here and in the PDF and compliance exports. The
existing reconciliation assertions are untouched; a new verbose fixture covers
the path the old one never exercised.

## Verification

- 9 + unit + endpoint tests confirmed **red first**, quoted above
- `ruff check src/ tests/` clean; `mypy` clean over 822 files
- every `make preflight` gate individually green
- touched suites green (126 + 148 + 6)
- after merging current `main`: `tests/cloud/test_cloud.py` +
  `test_extra_gated_sdk_imports.py` — 120 passed (the `openai._models` failures
  seen during development were the stub leak fixed by #4641)

**Not claimed:** the full suite was never run cleanly against the final state.
The run that completed (3 failed / 18841 passed) predates the MCP/CLI
consolidation and was taken across two `git stash push/pop` cycles, so it may
have executed mixed code. Treat it as evidence the tree is broadly healthy, not
as a green gate.

## Deliberately out of scope

Parquet, OCSF and the Iceberg catalog do not carry the verdict. Their column
schema is versioned (`_SCHEMA_VERSION` v2) and adding columns is a
consumer-visible bump — P2, next release.

No UI gates were run (`ui/node_modules` absent). The API change is additive and
`total_approximate` / `facets_approximate` only became more accurate, so no UI
change is believed necessary — reasoned, not executed.

## Non-finding, recorded

`print_posture_summary` counts CVEs by `effective_severity` but policy findings
by raw `severity`. No failing case could be constructed: nothing in `src/` ever
assigns `vendor_severity` or `cvss_severity` on a non-CVE finding, so the two are
always equal. Latent style divergence only; unchanged.


msaad00:
Closing the two gates the description left open.

**Full suite, clean run against the final state:** 3 failed, 18843 passed, 112 skipped (20m03s). This run started after every code change with no concurrent git operations, so it supersedes the stash-polluted run the description referred to.

All three failures are environmental and unrelated:
- 2x `test_extra_gated_sdk_imports` — `ModuleNotFoundError: No module named 'openai._models'`. This is the stub leak fixed by #4641; after merging current `main` into this branch, `tests/cloud/test_cloud.py` + `test_extra_gated_sdk_imports.py` give **120 passed**.
- 1x `test_online_scan_validation` — cannot pull `alpine:3.14.2`, no registry access.

**`make preflight` re-run against the final state — passed.** The description noted it had only been verified against an earlier state, before the MCP/CLI `verify_packages` consolidation and a `docs/SUPPLY_CHAIN.md` edit. Re-run now on the merged branch, all gates green including the product-metrics snapshot.

Both caveats in the description are therefore resolved; the rest of it stands.

andres-linero (approved):

agent-bom (approved):

PR #4648, opened 2026-08-04
fix: make the five persona promises actually deliver

### LINKED ISSUE #292: feat: transitive dependency graph visualization
## Summary

Visualize transitive dependency chains so users can see how a vulnerable deep dependency reaches their MCP servers.

## Context

`src/agent_bom/transitive.py` already resolves transitive deps via deps.dev API. This issue adds visualization.

## Scope

- Generate dependency tree (text format for CLI, JSON for MCP tool)
- Show path: MCP server → direct dep → transitive dep → vulnerable package
- SVG output option (extend existing SVG rendering in `output/`)
- Highlight vulnerable paths in red
- Add `--show-deps` flag to `scan` and `check` commands

msaad00:
## Summary

Walked each README persona's journey end to end with real commands, not code
reading. Four of five promises hold; **GRC was broken** — three of the sixteen
advertised compliance surfaces could not produce evidence at all, and the default
signed bundle was unverifiable.

13 defects fixed, 4 of them P0.

## Per-persona verdicts

| Persona | Verdict |
|---|---|
| **Developers** — accurate SCA, 15 ecosystems, EPSS/KEV, distro-aware | **HOLDS** (was partial) |
| **AppSec** — reachability, blast radius, CI gates, exit codes | **HOLDS**, one scope limit |
| **Platform / SRE** — self-hosted control plane, Postgres, signed audit | **HOLDS** |
| **GRC / audit** — audit-ready exports, control mappings, signed bundles | **WAS BROKEN**, now holds |
| **AI / MCP owners** — 380 API ops, 77 MCP tools, SARIF | **PARTIAL** |

Numbers verified exactly: live `/openapi.json` = 321 paths / **380 operations**;
live `tools/list` = **77**. Distro-awareness proven on a synthetic Alpine rootfs —
all 25 findings `distro_confirmed`, purl `pkg:apk/alpine/openssl@1.1.1k-r0`, and
the **Alpine-specific** fix `1.1.1l-r0` rather than the upstream version.

## The P0s

**1. PCI DSS was never tagged by any scan.** `pci_dss.tag_blast_radius` had zero
production callers while `package_scan.py` stamped 14 of 15 frameworks. A
framework counted in "16 compliance surfaces" could only ever report
`not_evaluated`.

```
RED: registered frameworks a scan never tags: ['pci_dss_tags']
```

Collected into `apply_framework_tags()` with a registry-driven test, so a new
framework that isn't stamped fails immediately. `not_evaluated → 9 failing`.

**2. FedRAMP could never score, and the bundle contradicted itself.** Tags emit
namespaced (`FedRAMP-SI-10`), the catalog is keyed bare, and only the exporter
stripped the prefix — so a signed bundle listed controls with mapped findings
while its own summary said `evaluated: 0, status: no_data`.
`not_evaluated → action_required, 15 failing`.

**3. The default evidence bundle was unverifiable, and the auditor endpoint gave
unfollowable advice.** Ephemeral per-process HMAC key, bundle stamped
`HMAC-SHA256` with no disclosure, and `/v1/compliance/verification-key` told
auditors to obtain a key that never left the process. Now returns
`signature_verifiable: false, verification_status: unverifiable_ephemeral_key`
with real remediation, and the disclosure sits **inside the signed envelope** —
proved by deleting it and getting `signature_mismatch` on a live bundle in a
fresh process. `agent-bom attest compliance-verify` ships as the first non-test
caller of `verify_compliance_bundle()`, refusing the bundle's self-attesting
embedded key by default.

**4. `scan(package="flask==0.12.2")` returned a false clean over MCP.** Any
unrecognized launcher was forced to `npx`, so a PyPI spec was scanned as npm: 0
packages, 0 warnings, presented as a complete AI-BOM — while `check(...)` on the
same input reported 4 CVEs. Now 4 CVEs; unresolvable specs fail closed with
`status: incomplete_scan`. Three more wrong answers of the same class were found
and fixed (`pip install X`, `uv tool run X`, `serde@1.0.0`), including in the
shared extractor a real MCP config goes through.

## Also fixed

- **`agent-bom connect` exited 0 on failure** and hid the reason. Now exits 1 (2
  for 400/422). The repro surfaced a third defect: every curated remediation
  (214/217/442 chars) was chopped mid-word by `sanitize_error`'s 200-char cap.
- **`POST /v1/scan {format: "sarif"}` silently returned JSON** — the enum was
  published in the live OpenAPI with zero consumers. All six formats now render;
  a test asserts the API Literal and `SCAN_DOCUMENT_FORMATS` are the same set.
- **`agent-bom scan .` produced nameless assets.** `Path(".").name` is `""`, so
  the README's headline command emitted `project:` and SARIF read
  `Exposure path: → → lodash@4.17.15`. This also explains the empty-named
  CycloneDX components.
- **The compliance narrative had no engineer altitude** — the shipped markdown
  rendered no control ids, packages or finding ids; `grep -c "CVE-"` returned 0,
  now 177.
- **`init.sql` hardcoded the database name.** Reproduced on real PostgreSQL 16.14
  with `POSTGRES_DB=abom_custom`: two `ERROR: database "agent_bom" does not exist`
  aborting the enclosing DO blocks.
- **README routed shared deployments to the insecure pilot** (loopback, SQLite,
  `--allow-insecure-no-auth`).
- Tracker references shipped inside the wheel — `agent-bom graph --help` printed
  "Closes #292." to every user. 6 sites cleaned, class guarded.

## Reported, not fixed

- **`warn` is not a real tier.** `GatewayRuntimeEventType` has no warn member and
  `proxy.py` calls the 2-tuple `check_policy`, so the proxy has **no warn tier at
  all** — a warned call is byte-identical to an allow. In the gateway, warn emits
  a non-durable alert plus `tool_call.allowed`, so KPIs count it as authorized.
  Needs a taxonomy decision, not a patch.
- **Blocked events lose rule attribution.** `gateway_server.py:2537` uses
  `check_policy` rather than `check_policy_detail` — whose docstring says it
  exists "so every block is traceable to a specific rule" — and the proxy route
  omits `rule_id` from the field allowlist. Both halves need fixing together.
- **Symbol reachability only fires from agent-framework tool entrypoints**
  (`ast_python_analysis.py:1717`). A plain Flask/Django app gets
  `reachability: unknown` on every finding. Honest, but narrower than the card
  implies.
- `--detect-credentials` redacts the AWS key ID but passes the 40-char secret.
- `/v1/gateway/stats` reports zeros while the gateway blocks (file-policy
  deployments). ATT&CK denominator differs by surface (413 vs 691) — a product
  call. Two Helm example profiles can't render as documented, and
  `validate_helm_profiles.py --list` excludes exactly those two.

## Verification

- `make preflight` ✓ — all gates including product-metrics and the data-model atlas
- `check-counts.py` ✓ — 321 REST paths, 77 MCP tools
- `mypy` clean on every changed source file; `ruff@0.9.7` (the pinned version)
  check + format clean on all changed files
- `tests/api` — **1189 passed** after merging current `main`
- Touched suites together — 1495 passed; `-k` sweeps over neighbours ~10k passing
- Postgres verified end to end on ephemeral PG16: 87 tables, 80 RLS policies,
  signed audit valid across a **process restart**, direct SQL tampering caught
  (`{"verified":4,"tampered":4}`)
- Helm: 19 profiles rendered with structural assertions (every Ingress rule has
  non-empty paths and a named service/port, every Deployment a pinned image and 3
  probes) — not merely "renders without error"
- SBOM passes the official CycloneDX 1.7 `JsonStrictValidator`, 0 dangling
  `dependsOn`

One environment-only failure throughout: `test_agent_bom_recall_matches_trivy_on_pinned_image`,
which is `network`-marked, excluded by CI's `-m "not network"`, and needs a
Docker daemon.


msaad00:
Correction to the verification section: it said "one environment-only failure throughout". That was based on the targeted suites, and it understates what a full run shows.

A whole-repo **single-process** run of this branch returned **20 failed / 18878 passed**. Two clusters were identified from the retained output:

- 6x `tests/cloud/test_cloud_gcp_inventory.py` — `module 'google.cloud.resourcemanager_v3' has no attribute 'FoldersClient'`. The installed SDK **does** export `FoldersClient`, and the file passes **45/45 in isolation**.
- 1x `tests/test_doc_architecture_svgs.py::test_all_generated_svg_bytes_match_their_generators` — passes **12/12 in isolation**, and preflight's own `generate_doc_architecture_svgs.py --check` passes.

Both sit in code this branch does not touch, and both reproduce only under full-suite ordering — the signature of cross-test pollution, a test stubbing `sys.modules` without restoring it. That is the same defect class as the `openai` stub leak fixed in #4641, where `_STUB_SDK_ROOTS` was missing an entry.

**That is a plausible explanation, not proof, and I am not claiming this branch is green on a full single-process run.** What the targeted evidence does support, and what was actually executed: `tests/api` 1189 passed after merging current `main`, all touched suites together 1495 passed, ~10k passing executions across neighbour sweeps, `make preflight` green, `mypy` and pinned `ruff@0.9.7` clean.

Worth separating out as its own item regardless of how this branch lands: **a full single-process run of this suite is not currently trustworthy.** CI invokes it sharded (`-n <workers> --dist worksteal`, `-m "not network and not slow"`), so this is invisible there — which is precisely why it has survived. Filing it against the branch would conflate an unrelated pre-existing problem with these persona fixes.

msaad00:
Resolved: **the branch suite is green.** Re-running the identical command on the same tree: **18928 passed, 111 skipped, 0 failed, exit 0**.

The difference between the two runs is `pytest-randomly`, which this repo enables deliberately (`pyproject.toml:251` — *"Randomize test order to surface order-dependence early"*). The first run drew an order that tripped 20 order-dependent tests; the second drew one that did not. So the suite has genuine order-dependence, and the repo's own randomization is what surfaced it — working as designed.

Not caused by this branch, established three ways: both clusters pass in isolation (GCP inventory 45/45, doc SVGs 12/12), neither sits in a file this branch touches, and a full clean pass on the identical tree rules out the branch as the cause.

**Mechanism — a strong lead, not a diagnosis.** `tests/test_cis_denied_read_coverage.py:133-143` assigns `sys.modules["google"]` and `sys.modules["google.cloud"]` to mocks by direct assignment and never restores them; `tests/test_gcp_cis_benchmark.py` does the same for `compute_v1`, `storage`, `logging_v2`. Neither uses `monkeypatch.setitem` or `del sys.modules` (0 occurrences in each). A later `import google.cloud.resourcemanager_v3` resolving against a bare stub produces exactly the observed `no attribute 'FoldersClient'`, and the real SDK does export it.

But running those files immediately before the GCP inventory suite **passes** (62/62, 116/116), so the trigger needs a specific interleaving that could not be reproduced — the first run's `--randomly-seed` was lost when the output file kept only its last lines. Reported as an unconfirmed hypothesis with a concrete lead. If it holds, the fix is `monkeypatch.setitem(sys.modules, ...)` so the stubs unwind — the same shape as #4641, which added `openai` to `_STUB_SDK_ROOTS` for exactly this reason.

**One correction, in the stricter direction.** My description listed `test_agent_bom_recall_matches_trivy_on_pinned_image` as an environment-only *failure*. It is `network`-marked and was excluded from both full runs, so it never executed — it is **unverified**, not passing.

Final gates: whole-repo `pytest -m "not network"` 18928 passed / 0 failed; `make preflight` pass; `check-counts.py` pass (321 paths, 77 tools); `mypy` clean; pinned `ruff@0.9.7` check and format clean.

msaad00:
Baseline settles it definitively.

| Tree | Result |
|---|---|
| `origin/main` @ `704e99682`, pristine `git archive` | **19 failed**, 18802 passed |
| this branch | **0 failed**, 18928 passed, exit 0 |

All 19 baseline failures are the identical `AttributeError: module 'google.cloud.resourcemanager_v3' has no attribute 'FoldersClient'`, and they take out **the entire `tests/cloud/test_cloud_gcp_inventory.py` file** — every test in it. So by the time that file runs, the module is poisoned for the rest of the process. The installed SDK genuinely exports `FoldersClient`; the file passes 45/45 alone.

So this is a **pre-existing defect on `main`**, not something this branch introduced — and roughly half of all full single-process runs on `main` are red for a reason unrelated to whatever change is under test. CI's sharded `-n workers --dist worksteal` splits tests across processes, which is very likely why it has stayed invisible.

Deliberately **not fixed here**: it is outside this branch's scope, it cannot yet be reproduced on demand, and a fix that cannot be made to fail first is the kind of guess TDD exists to prevent. Tracked separately; suggested **P1** — it ships no wrong output to users, but it corrodes trust in every full-suite run and can mask a real regression behind noise.

The lead remains the unrestored `sys.modules` writes described in the previous comment.

msaad00:
Two corrections, both against my own earlier comments.

**1. I overstated the mechanism.** I called the unrestored `sys.modules` writes a "strong lead". They are not — three bounded attempts failed to reproduce the failure on demand:

- the two suspect files run immediately before `test_cloud_gcp_inventory.py` pass (62/62, 116/116)
- all 1415 tests across every `google`-touching file, in deterministic order, pass in 34s

So the hypothesis is **unconfirmed and unsupported by the attempts made to confirm it.** The genuinely useful result is the negative one: `"google"` and `"google.cloud"` are **already** in `_STUB_SDK_ROOTS` in `tests/conftest.py`, so the existing snapshot/restore did not prevent this. The obvious explanation is wrong, which is worth knowing before someone spends an afternoon on it. The mechanism should be recorded as **unknown**.

What remains solid is the phenomenon and that it predates this branch:

| Run | Result |
|---|---|
| `origin/main` branch point, pristine `git archive` | **19 failed** — the entire `test_cloud_gcp_inventory.py` file |
| this branch, run 1 | 20 failed |
| this branch, runs 2 and 3 | **0 failed** (18928, 18946) |

Same tree, opposite outcomes, because `pytest-randomly` reshuffles each run.

**2. The merge commit `81abc27bb` on this branch is mine**, not an unexplained mutation. I merged `origin/main` to resolve three conflicts — `docs/DATA_MODEL.md` and `docs/PRODUCT_METRICS.{json,md}`, all generated count files — by taking main's side and regenerating via `product_metrics_snapshot.py --write` and `regenerate_data_model_atlas.py`. `make preflight` and `tests/api` (1189 passed) were run on the merged result before pushing.

That merge does mean the gates reported against `68bbba557` were not the merged tree. Re-run on merged HEAD: `make preflight` pass, touched suites **1495 passed / 1 skipped**, `check-counts` pass (321 paths, 77 tools), `mypy` clean. A whole-repo deterministic run was still in flight.

agent-bom (dismissed):

andres-linero (dismissed):

agent-bom (approved):

andres-linero (approved):

PR #4656, opened 2026-08-04
fix(gcp): degrade on an incomplete SDK, and stop the fake SDK leaking real modules

msaad00:
## Summary

Chasing an order-dependent suite failure found a **production defect**: a
partially-installed `[gcp]` extra crashes the inventory scan instead of
degrading with the warning that exists for exactly that case.

## The production bug

`gcp_authorization_collector._load_clients` (`:199`) imports five sibling
modules and builds seven clients. The caller guards it with `except ImportError`
(`:253`) and, on that path, warns *"GCP IAM evidence SDKs are incomplete. Install
with: pip install 'agent-bom[gcp]'"* and returns empty evidence.

That guard cannot catch the case where the module imports fine but a **client
class is missing** — a genuinely partial SDK install. Then `_load_clients` raises
`AttributeError: module 'google.cloud.resourcemanager_v3' has no attribute
'FoldersClient'` at `:211`, which escapes the `except ImportError` and the
unguarded call site at `gcp_inventory.py:431`, taking down the whole scan.

Crash-instead-of-degrade, against the project's own graceful-degradation rule.
`_load_clients` now converts a missing client class into `ImportError`, so the
documented path runs.

## Why the suite was flaky

`google.cloud` is a **PEP 420 namespace package**. The test fake
(`_install_fake_gcp`) installs a bare `types.ModuleType("google.cloud")` as the
parent, which has no `__path__` — so `_handle_fromlist` skips it and
`from google.cloud import asset_v1` is resolved **solely** by CPython's
`IMPORT_FROM` fallback to `sys.modules["google.cloud.asset_v1"]`.

Whatever an earlier test left resident there is handed back. The fake stubs only
`resourcemanager_v3` and `iam_admin_v1`, so depending on what ran first the
import either raised `ImportError` and degraded (green), or bound the **real**
`asset_v1`/`iam_v2`/`iam_v3` and walked into the fake `_FakeResourceManagerModule`,
which defines only `ProjectsClient` — hitting the production bug above.

That is why the same tree passed or failed depending on `pytest-randomly`'s draw.

Deterministic reproduction, no seed required — pre-import the real siblings and
run the victim file alone: **exactly 19 failed, 26 passed**, matching a pristine
`origin/main`. After the fix: **47 passed**.

## Why `_STUB_SDK_ROOTS` could never have fixed it

`"google"` and `"google.cloud"` are already listed in `tests/conftest.py`, and
the restore still did not prevent this — which is what made the obvious fix
wrong. `_restore_stub_sdk_modules` only deletes or replaces entries that
`_is_stub_module` judges to be stubs, and the leaking modules are genuinely
imported site-packages modules with a real `__spec__`, deliberately excluded
(`conftest.py:658-659`).

The leak is not a stub outliving its test. It is a **real module outliving its
import**, which is ordinary Python. No snapshot/restore policy fixes that — only
the stub site can, so `tests/_sdk_stub_helpers.patch_sdk_namespace` removes every
other resident module under the stubbed roots.

## Verification

- Three tests, all confirmed **red first**
- `tests/cloud/test_cloud_gcp_inventory.py` — 47 passed (was 19 failed / 26
  passed under the reproducing condition)
- Touched suites — 842 passed
- `ruff check`, `ruff format --check`, `mypy` — clean
- `make preflight` — passed, re-run after merging current `main`

## Not claimed

The **in-suite** poisoner is not named. A full-suite detector reached 70% with
zero hits, which suggests the poisoner is itself order-dependent. The mechanism
above is proven by deterministic reproduction rather than by identifying the
specific interleaving, and the fix addresses the mechanism, not a particular
ordering. Full-suite verification runs were still in flight at report time.

### DIFF
--- a/src/agent_bom/cloud/gcp_authorization_collector.py
+++ b/src/agent_bom/cloud/gcp_authorization_collector.py
@@ -196,6 +196,26 @@ def _pab_binding_record(binding: Any) -> dict[str, Any]:
     }
 
 
+def _client_class(module: Any, attribute: str) -> Any:
+    """Return ``module.attribute``, or raise ``ImportError`` when the SDK lacks it.
+
+    ``google.cloud`` is a namespace package assembled from one distribution per
+    service, so the modules can import while still missing the client classes we
+    need — a partial ``[gcp]`` install, or a version predating a client. That is
+    the same operator condition as an absent SDK, so it is reported the same way:
+    the caller's ``ImportError`` handler degrades to ``SDK_MISSING``. Letting the
+    raw ``AttributeError`` escape would abort the whole GCP inventory walk, whose
+    call site does not guard it.
+    """
+    try:
+        return getattr(module, attribute)
+    except AttributeError as exc:
+        raise ImportError(
+            f"{getattr(module, '__name__', module)} has no {attribute} — the installed "
+            "google-cloud SDKs are incomplete. Install with: pip install 'agent-bom[gcp]'"
+        ) from exc
+
+
 def _load_clients(credentials: Any) -> Any:
     from google.cloud import (  # type: ignore[attr-defined]  # namespace package exports
         asset_v1,
@@ -206,14 +226,14 @@ def _load_clients(credentials: Any) -> Any:
     )
 
     return SimpleNamespace(
-        assets=asset_v1.AssetServiceClient(credentials=credentials),
-        projects=resourcemanager_v3.ProjectsClient(credentials=credentials),
-        folders=resourcemanager_v3.FoldersClient(credentials=credentials),
-        organizations=resourcemanager_v3.OrganizationsClient(credentials=credentials),
-        roles=iam_admin_v1.IAMClient(credentials=credentials),
-        denies=iam_v2.PoliciesClient(credentials=credentials),
-        pabs=iam_v3.PrincipalAccessBoundaryPoliciesClient(credentials=credentials),
-        policy_bindings=iam_v3.PolicyBindingsClient(credentials=credentials),
+        assets=_client_class(asset_v1, "AssetServiceClient")(credentials=credentials),
+        projects=_client_class(resourcemanager_v3, "ProjectsClient")(credentials=credentials),
+        folders=_client_class(resourcemanager_v3, "FoldersClient")(credentials=credentials),
+        organizations=_client_class(resourcemanager_v3, "OrganizationsClient")(credentials=credentials),
+        roles=_client_class(iam_admin_v1, "IAMClient")(credentials=credentials),
+        denies=_client_class(iam_v2, "PoliciesClient")(credentials=credentials),
+        pabs=_client_class(iam_v3, "PrincipalAccessBoundaryPoliciesClient")(credentials=credentials),
+        policy_bindings=_client_class(iam_v3, "PolicyBindingsClient")(credentials=credentials),
     )
 
 
--- a/tests/_sdk_stub_helpers.py
+++ b/tests/_sdk_stub_helpers.py
@@ -1,0 +1,52 @@
+"""Install a fake provider SDK in ``sys.modules`` without leaking the real one.
+
+``patch.dict(sys.modules, fakes)`` shadows only the names it lists, which is not
+enough for a namespace package like ``google.cloud``. A stub parent is a bare
+``types.ModuleType`` with no ``__path__``, so ``_handle_fromlist`` skips it
+entirely and ``from google.cloud import asset_v1`` is resolved solely by
+CPython's ``IMPORT_FROM`` fallback to ``sys.modules["google.cloud.asset_v1"]``.
+Whatever an earlier test in the same process left resident there is handed back —
+so a test that stubs only part of a namespace silently binds real SDK modules,
+and its outcome depends on which tests ran before it.
+
+Mapping the extra names to ``None`` does not close the hole either: the same
+fallback returns the ``None`` object rather than raising, and the caller then
+fails on ``None.SomeClient``. The entries have to be removed outright, which is
+what :func:`patch_sdk_namespace` does — and restores on exit.
+"""
+
+from __future__ import annotations
+
+import sys
+from collections.abc import Iterator
+from contextlib import contextmanager
+from typing import Any
+
+
+@contextmanager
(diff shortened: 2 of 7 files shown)


msaad00:
A **second, latent leak** was found after this PR opened, and is now included (`6ef3eaec2`).

`_ensure_google_namespace` (`test_gcp_cis_benchmark.py`) and `_install_mock_gcp_storage` (`test_cis_denied_read_coverage.py`) reused the **real** `google.cloud` package and bound fakes onto it as *attributes*. conftest restores `sys.modules`, but it cannot undo an attribute set on a package object — and attribute lookup wins before the import machinery runs.

So after either file executed, the real `google.cloud` carried `storage`, `compute_v1` and `logging_v2` permanently rebound to test doubles for the rest of the process. A session-end probe on pristine `main` found three fake attributes still bound; after the fix, none.

This sharpens the point about `_STUB_SDK_ROOTS`: it could not have caught **either** defect, for two different reasons.

- Defect 1 is not a stub outliving its test — it is a *real module* outliving its import, which is ordinary Python. The bug is on the consumer side, in production code that assumed `except ImportError` covered a partial SDK.
- Defect 2 is not a `sys.modules` entry **at all**. It is a package attribute, which no `sys.modules` snapshot/restore policy can reach.

Gates: touched suites green (842 + 101 + 89), `ruff` and `mypy` clean, `make preflight` green twice, and a full **deterministic single-process** run of the fixed tree at **18996 passed, 110 skipped**.

Unchanged caveat, stated plainly: an instrumented full run found **zero** tests leaving the real siblings resident, so the in-suite poisoner and its minimal interleaving are still unidentified — the trigger appears order-dependent itself. Both fixes address proven mechanisms, each with a deterministic reproduction, rather than a particular ordering. If no naturally-red seed turns up on pristine `main`, that will be reported as unresolved rather than invented.

andres-linero (approved):

agent-bom (approved):

msaad00:
**Correction, and then the gate evidence.**

My earlier comment said *"an instrumented full run found zero tests leaving the real siblings resident."* **That result was invalid.** The detector `print`ed inside a pytest runtest hook, where stdout is captured and discarded under `-q` — so it was reporting nothing found because it could report nothing at all. It now writes to a file and has been re-verified against a known subset.

The mistake cost time and changed no conclusion: the root cause is confirmed by a real traceback, not by that detector.

**Gate results:**

| Run | Tree | Result |
|---|---|---|
| seed 1 | pristine `origin/main` | **19 failed**, 18983 passed — reproduces the report exactly |
| seed 1 | this branch | **0 `FoldersClient` failures**, 19009 passed |
| seed 2001 | this branch | **RC=0**, 19024 passed, clean |

So the phenomenon now has a deterministic reproduction (`--randomly-seed=1` on pristine main) and the fix is verified against that exact seed.

The pristine seed-1 traceback confirms the mechanism precisely: `asset_v1.AssetServiceClient(credentials=None)` **constructs successfully** — proving `asset_v1` resolved to the *real* module — and then `resourcemanager_v3.FoldersClient` raises against the test's fake, at `gcp_authorization_collector.py:211` via `gcp_inventory.py:431`. That is the partial-namespace collision this PR describes, observed rather than inferred.

Two caveats I will not paper over:
- The seed-1 replay on the fix carried 5 **environmental** errors, none GCP: three `OSError: [Errno 48] Address already in use` in `test_http_client_proxy_env.py` from concurrently running suites, and two kind/`kubectl` errors that also appear on the pristine run. That replay is being redone serially so the result is clean rather than argued.
- **The in-suite poisoner is still unnamed.** The re-instrumented detector is running. The fix targets the mechanism, which is reproducible on demand; it does not depend on identifying which test triggers it.

msaad00:
**The poisoner is named, and the repro is now two files in 17 seconds.**

`tests/test_extra_gated_sdk_imports.py::test_extra_gated_sdk_import_executes` scrapes every extra-gated SDK import out of `src/agent_bom` and **executes** it — including `from google.cloud import asset_v1` at `gcp_authorization_collector.py:200`. Executing the import *is* the check, so it legitimately leaves the real `google.cloud.asset_v1`, `iam_v2` and `iam_v3` resident for the rest of the session. **Nothing is wrong with that test.**

Minimal interleaving on pristine `origin/main`:

```
pytest -p no:randomly tests/test_extra_gated_sdk_imports.py tests/cloud/test_cloud_gcp_inventory.py
  → 19 failed, 133 passed      # poisoner first

pytest -p no:randomly tests/cloud/test_cloud_gcp_inventory.py tests/test_extra_gated_sdk_imports.py
  → 152 passed                  # victim first
```

On this branch, **both orders: 154 passed**.

That also explains the ~50% failure rate exactly. Alphabetically `tests/cloud/` sorts before `tests/test_extra_*`, so a deterministic run is *always* green — `pytest-randomly` shuffles file order, so roughly half of runs put the poisoner first. The suite was only ever red by coin flip, which is why it survived.

It also confirms why the fix belongs where it is: the poisoner is correct, the victim's partial namespace stub is the defect, and the production code's `except ImportError` was never going to catch an `AttributeError` from a real-but-partial SDK.

**One disclosure about the evidence.** Another process — a concurrent agent session, not this work — operated inside the branch's worktree mid-run, merging `origin/main` and checking out other branches. My branch consequently carries merge commit `272a26732` that the fix's author did not write; its content is just `origin/main`, the 8 changed files are intact, and the tree is clean.

The consequence for the gate: **branch seed 2001 ran underneath that churn and is being discarded rather than counted.** Seed 2002 (RC=0, 19024 passed) started afterwards and stands. A clean uncontended seed-1 replay plus one more pass and a preflight re-run are in flight before the gate is called met.

msaad00:
**The replayed failing seed passes on this branch, run uncontended.**

| | pristine `origin/main` | this branch |
|---|---|---|
| `--randomly-seed=1` | **19 failed**, 18983 passed | **RC=0**, 19014 passed, 105 skipped |
| `FoldersClient` errors | 19 | **0** |

Same seed, same interpreter, same tree except the fix. No failures and no errors — the port-collision and `kind` noise in the earlier attempt came from running it alongside other suites and is gone now that it ran alone.

That closes the gate item: the exact ordering that reproduced the report is now green. Combined with the two-file 17-second repro and both file orders passing on the fix, the mechanism is demonstrated rather than argued.

Second clean full pass (seed 2004) and a preflight re-run on final HEAD are in flight; those are confirmation, not the gate.

PR #4659, opened 2026-08-04
perf: stop blocking the loop and re-deriving work at estate scale

msaad00:
## Measured first

A prior optimisation attempt on the generator guessed twice and was wrong both
times — `model_copy` was assumed to be the bottleneck (1.98s → 1.93s, reverted)
and threading was assumed to help (**3× slower**, GIL-bound). So this profiled
before touching anything.

`/v1/graph?limit=5000` at 2,068 nodes, ms/request:

| item | ms/req | share |
|---|---|---|
| `_hashlib.scrypt` (API-key verify) | 43.0 | 31% |
| `zlib.compress` (gzip) | 31.2 | 23% |
| `page_nodes` (query + row build) | 10.2 | 7% |
| sqlite `execute`/`fetchall` | 9.2 | 7% |

**The two largest items were not graph work at all.**

## The biggest finding was outside the routes

`/v1/demo-estate/story` — the one route serving the composed estate — took
**1,643 ms per request with the event loop parked for 1,648 ms of it**: a
synchronous build inside an `async def`, uncached. Four concurrent callers
serialised into **6,630 ms** on one worker.

Now **28 ms warm**, loop stall **1 ms**, four concurrent **47 ms** — 141×.
Cacheability was proven before caching: identical `story_content_hash` across
runs, frozen model, no cross-tenant leakage. The cache is keyed per tenant (the
payload embeds tenant-specific hashes, so a shared entry would be *disclosure*,
not staleness) and bounded at 8 entries (~1.5 MB each).

## Fixes, each with measured before/after

**Redundant scrypt on every authenticated request.**
`RateLimitMiddleware._resolve_tenant_scope` re-verified the key
`APIKeyMiddleware` had already verified, and its short-circuit excluded tenant
`"default"` — exactly the single-tenant and hosted-demo case. Verified
end-to-end, not just in a unit test: **2 → 1 derivations/request, 21.5 ms saved
on every route in the API.**

**`page_nodes` sorted the whole snapshot per page** — `USE TEMP B-TREE FOR ORDER
BY`. Added the matching composite index:

| nodes | before | after | speedup |
|---|---|---|---|
| 2,068 | 0.69 ms | 0.26 ms | 2.6× |
| 100,000 | 11.84 ms | 0.48 ms | 24.6× |
| 300,000 | 31.83 ms | 0.53 ms | **59.6×** |

Unindexed cost grows linearly, indexed stays flat. This is the 444ms@300k class
from the audit history — **Postgres already had this index; the SQLite store had
drifted behind it.**

**GZip at Starlette's default level 9**, compressing synchronously on the loop.
On the 2.7 MB payload: 32.5 ms/175.4 KB at level 9 vs **17.7 ms/185.2 KB at
level 6**. Halved loop time for 5.6% more bytes.

Controlled A/B, same script back-to-back, files swapped via `git checkout`:
every route improved **26–45%**.

## Already fast — negative results, no change made

- **First-page exact `COUNT(*)` is not linear** — covering index, 0.040 ms.
- **Keyset cursors do not truncate to page 1** — full walk returned all 2,068
  ids across 7 pages, zero dupes, zero drops, re-verified after the index change.
- **Node-budget truncation is visible** (#4631 holds) — `completeness` reports
  `truncated/returned/total` with a reason.
- **`cvss` filesort is not applicable** — `graph_nodes` has no cvss column; that
  defect lives in the findings read path.
- **`canonical_id`/`uuid5` (5% of the profile)** — deliberately not memoised:
  `UnifiedNode` is mutable, so caching risks staleness for a 5% gain.

## An adversarial catch

The first offload created a new problem: anyio's default thread limiter is **40
tokens and shared**, and API-key scrypt uses the same pool. A burst of cold story
requests would hold a thread each while queueing on the build lock and **starve
authentication**. A 2-token `CapacityLimiter` (second commit) makes surplus wait
on the loop instead of on a worker thread.

## Verification

- `pytest -k "graph or scope or pagination or read or perf"` — 2169 passed
- Re-verified after merging `main`: 1348 passed, 29 skipped
- `mypy`, `ruff check`, `ruff format --check` — clean; `make preflight` — passed
- New tests confirmed to fail without the fixes (2,108 ms stall; 3 builds instead
  of 1). The event-loop test drives the ASGI app on the **same** loop as a
  heartbeat, because `TestClient` uses its own loop where a blocking route is
  invisible.

## Flagged, not fixed

- **Cloud Run behaviour is extrapolated, not measured** — all numbers are from an
  M-series box. Loop and scrypt costs should be *worse* on a modest vCPU.
- **`build_demo_estate()` is not in the graph.** The demo seeds `showcase_graph`:
  **112 nodes / 142 edges**. The 2,068-asset estate reaches only the story route
  and the bootstrap fingerprint. The 2,068-node graph measured here is a
  synthetic projection written for the measurement, not a shipped path. Either
  project the estate into the graph, or the "correlated graph at enterprise
  scale" claim rests on 112 nodes.
- **Measured but deliberately not shipped:** `security.SENSITIVE_PATTERNS` is a
  list of raw strings hit via `re.search` — 4.19M `re._compile` cache lookups in
  the story profile. One precompiled alternation is 10.6× faster with no
  predicate mismatches on the sample. Security-critical redaction deserves its
  own change with a proper equivalence corpus, not a rider on a perf PR.

### DIFF
--- a/docs/PRODUCT_METRICS.json
+++ b/docs/PRODUCT_METRICS.json
@@ -28,7 +28,7 @@
     },
     {
       "name": "Test files",
-      "value": 1132,
+      "value": 1133,
       "source": "tests/",
       "notes": "Counts files matching test_*.py."
     },
--- a/docs/PRODUCT_METRICS.json
+++ b/docs/PRODUCT_METRICS.json
@@ -28,7 +28,7 @@
     },
     {
       "name": "Test files",
-      "value": 1131,
+      "value": 1132,
       "source": "tests/",
       "notes": "Counts files matching test_*.py."
     },
@@ -46,7 +46,7 @@
     },
     {
       "name": "Python modules",
-      "value": 829,
+      "value": 830,
       "source": "src/agent_bom",
       "notes": "Counts all Python files recursively."
     },
--- a/docs/PRODUCT_METRICS.md
+++ b/docs/PRODUCT_METRICS.md
@@ -14,10 +14,10 @@ Keep counts out of public positioning copy and update this file from the repo in
 | MCP resources | 6 | `src/agent_bom/mcp_server_metadata.py` | Counted from the advertised server-card resources. |
 | MCP prompts | 8 | `src/agent_bom/mcp_server_metadata.py` | Counted from the advertised server-card workflow prompts. |
 | GitHub workflow files | 38 | `.github/workflows` | Counts .yml and .yaml workflow definitions. |
-| Test files | 1131 | `tests/` | Counts files matching test_*.py. |
+| Test files | 1132 | `tests/` | Counts files matching test_*.py. |
 | API route modules | 46 | `src/agent_bom/api/routes` | Counts Python files in the routes package, including __init__.py. |
 | UI app pages | 41 | `ui/app` | Counts page.tsx and page.jsx files recursively. |
-| Python modules | 829 | `src/agent_bom` | Counts all Python files recursively. |
+| Python modules | 830 | `src/agent_bom` | Counts all Python files recursively. |
 | Supported package ecosystems | 15 | `src/agent_bom/ecosystems.py` | Counted from SUPPORTED_PACKAGE_ECOSYSTEMS. |
 | Compliance surfaces | 16 | `src/agent_bom/compliance_coverage.py` | 15 tag-mapped frameworks plus the OWASP AISVS benchmark surface. |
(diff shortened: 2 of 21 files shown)


msaad00:
Superseded by #4660, which combines this with the other two and re-runs the full gate on the **merged** result — 18951 passed, 0 failed, plus mypy (831 files), ruff and preflight. The three branches merged with zero conflicts; the combined run is what proves no regression, since none of them ran the others' tests.

PR #4660, opened 2026-08-04
perf: stop blocking the event loop and re-deriving work at estate scale

msaad00:
Profiled before touching anything, because a prior optimisation attempt guessed
twice and was wrong both times — `model_copy` was assumed to be the bottleneck
(1.98s → 1.93s, reverted) and threading was assumed to help (**3× slower**,
GIL-bound).

At 2,068 nodes on `/v1/graph?limit=5000`, **the top two costs were not graph work
at all**:

| item | ms/req | share |
|---|---|---|
| `_hashlib.scrypt` (API-key verify) | 43.0 | 31% |
| `zlib.compress` (gzip) | 31.2 | 23% |
| `page_nodes` (query + row build) | 10.2 | 7% |
| sqlite `execute`/`fetchall` | 9.2 | 7% |

## The biggest finding was outside the routes profiled

`/v1/demo-estate/story` — the one route serving the composed estate — took
**1,643 ms per request with the event loop parked for 1,648 ms of it**: a
synchronous build inside an `async def`, uncached. Four concurrent callers
serialised into **6,630 ms** on one worker.

Now **28 ms warm**, loop stall **1 ms**, four concurrent **47 ms**.

Cacheability was proven before caching: identical `story_content_hash` across
runs, frozen model, no cross-tenant leakage. The cache is keyed **per tenant** —
the payload embeds tenant-specific hashes, so a shared entry would be
*disclosure*, not staleness — and bounded at 8 entries (~1.5 MB each).

## Fixes, each measured

**Redundant scrypt on every authenticated request.**
`RateLimitMiddleware._resolve_tenant_scope` re-verified the key
`APIKeyMiddleware` had already verified, and its short-circuit excluded tenant
`"default"` — exactly the single-tenant and hosted-demo case. Verified
end-to-end, including middleware order: **2 → 1 derivations/request, 21.5 ms
saved on every route in the API.**

**`page_nodes` sorted the whole snapshot per page** (`USE TEMP B-TREE FOR ORDER
BY`). Added the matching composite index:

| nodes | before | after | speedup |
|---|---|---|---|
| 2,068 | 0.69 ms | 0.26 ms | 2.6× |
| 100,000 | 11.84 ms | 0.48 ms | 24.6× |
| 300,000 | 31.83 ms | 0.53 ms | **59.6×** |

Unindexed cost grows linearly, indexed stays flat. **Postgres already had this
index; the SQLite store had drifted behind it** — the same one-thing-two-
implementations shape this round keeps finding.

**GZip at Starlette's default level 9**, compressing synchronously on the loop.
On the 2.7 MB payload: 32.5 ms/175.4 KB at level 9 vs **17.7 ms/185.2 KB at
level 6**. Halved loop time for 5.6% more bytes.

Controlled A/B, same script back-to-back with files swapped via `git checkout`:
every route improved **26–45%**.

## An adversarial catch, before it shipped

The first offload created a worse problem: anyio's default thread limiter is
**40 tokens and shared**, and API-key scrypt uses the same pool. A burst of cold
story requests would hold a thread each while queueing on the build lock and
**starve authentication**. A 2-token `CapacityLimiter` makes surplus wait on the
loop instead of on a worker thread.

## Already fast — negative results, deliberately unchanged

- **First-page exact `COUNT(*)` is not linear** — covering index, 0.040 ms.
- **Keyset cursors do not truncate to page 1** — full walk returned all 2,068 ids
  across 7 pages, zero dupes, zero drops, re-verified after the index change.
- **Node-budget truncation stays visible** (#4631 holds) — `completeness` reports
  `truncated/returned/total` with a reason.
- **`cvss` filesort does not apply** — `graph_nodes` has no cvss column; that
  defect lives in the findings read path.
- **`canonical_id`/`uuid5` (5% of the profile)** — not memoised: `UnifiedNode` is
  mutable, so caching risks staleness for a 5% gain.

## Verification

This branch began as a combine of #4657, #4658 and #4659. Those first two were
merged to `main` independently while the combined tree was being gated, so **the
diff here is the perf work alone** — but the gate ran on a tree containing all
three, which is exactly the combination now on `main`:

| Gate | Result |
|---|---|
| `pytest -m "not network" -p no:randomly` | **18951 passed, 191 skipped, 0 failed** |
| `mypy src/` | no issues, 831 source files |
| `ruff check src/ tests/ scripts/` | all checks passed |
| `make preflight` | passed, re-run after resolving the merge |

New tests were confirmed to fail without the fixes (2,108 ms stall; 3 builds
instead of 1). The event-loop test drives the ASGI app on the **same** loop as a
heartbeat, because `TestClient` uses its own loop where a blocking route is
invisible.

Deterministic order was used deliberately: after #4656, a random-order failure
here would report on ordering rather than on this change.

**One disclosure.** An earlier run of that suite reported 19 failed / 42 errors.
It passed through a disk-full window (`could not create numbered dir`) and is
discarded — it proved nothing. The result above is a clean re-run.

## Not covered

- **UI gates** (`tsc`, `vitest`, `build`) were not run on the merged tree —
  `npm ci` is what exhausted the disk. No UI files remain in this diff.
- **Cloud Run behaviour is extrapolated, not measured.** All numbers come from an
  M-series box; loop and scrypt costs should be *worse* on a modest vCPU.
- **Measured but deliberately not shipped:** `security.SENSITIVE_PATTERNS` is a
  list of raw strings hit via `re.search` — 4.19M `re._compile` cache lookups in
  the story profile. One precompiled alternation is 10.6× faster with no
  predicate mismatches on the sample. Security-critical redaction deserves its
  own change with a proper equivalence corpus, not a rider on a perf PR.

## Still open after this

`build_demo_estate()` is not projected into the graph — it renders **112 nodes /
142 edges** from `showcase_graph` while the estate holds 2,068 assets. The
2,068-node graph in these measurements is a synthetic projection written for the
benchmark, **not a shipped path**. That is the last gap between the evidence
layer and what a prospect actually clicks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,6 @@
 <p align="center"><b>Open security scanner and self-hosted control plane for AI, MCP, and cloud infrastructure.</b></p>
 
 <p align="center">
-  Scan repositories, images, and cloud accounts; centralize evidence across environments; and enforce AI and MCP runtime policy in infrastructure you control.
-</p>
-
-<p align="center">
   <b>15</b> package ecosystems · <b>16</b> compliance surfaces · <b>77</b> MCP tools · no account required<br />
   <a href="#quick-start"><b>Quick start</b></a> ·
   <a href="https://agent-bom-demo-82102570041.us-central1.run.app">Live demo</a> ·
@@ -32,19 +28,17 @@
 
 ## What it is
 
-`agent-bom` is an open scanner and self-hosted control plane for software,
-cloud, identity, AI-agent, and MCP evidence. One Finding + UnifiedGraph model
-powers CLI and CI artifacts, fleet and browser investigations, compliance
-evidence, and runtime policy.
+`agent-bom` scans repositories, images, and cloud accounts, then correlates what
+it finds into one Finding + UnifiedGraph model — powering CLI and CI artifacts,
+fleet and browser investigations, compliance evidence, and runtime policy. Run
+the scanner without an account, or deploy the control plane inside your own
+cloud, VPC, cluster, database, identity, and audit boundary.
 
-Use the scanner without an account, or deploy the shared control plane inside
-your own cloud, VPC, Kubernetes cluster, database, identity, and audit boundary.
-
-Graph provenance remains explicit: collected, inferred, static, and runtime
-relationships stay distinct, and unavailable evidence is never upgraded to
+Graph provenance stays explicit: collected, inferred, static, and runtime
+relationships remain distinct, and unavailable evidence is never upgraded to
 observed.
 
-<details>
+<details open>
 <summary><b>Control-plane architecture</b></summary>
 
 <p align="center">
--- a/docs/DATA_MODEL.md
+++ b/docs/DATA_MODEL.md
@@ -54,7 +54,7 @@ is wired into the docs site so drift produces a visible regression.
 | `config/schemas/inventory.schema.json` | `Package.ecosystem` enum values | 9 |
 | `config/schemas/inventory.schema.json` | `MCPServer.transport` enum values | 3 |
 | `docs/openapi/v1.json` | paths | 322 |
-| `docs/openapi/v1.json` | component schemas | 133 |
+| `docs/openapi/v1.json` | component schemas | 135 |
 
 <!-- DATA_MODEL_ATLAS:END -->
 
--- a/docs/ENTERPRISE_DEMO.md
+++ b/docs/ENTERPRISE_DEMO.md
@@ -31,6 +31,37 @@ enforced multi-provider remediation. GCP remains `partial` with
 `rate_limited_after_page_2`; the demo never turns missing evidence into a
 complete posture claim.
 
+## Posture findings over the estate
+
+The estate also carries posture findings, so the correlation is demonstrated at
+estate scale rather than only along the hand-authored incident. Each finding
+resolves to a chain:
+
+```text
+finding → inventoried asset → identity that can act on it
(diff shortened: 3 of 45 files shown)


msaad00:
**Scope correction.** This opened as a three-way combine of #4657, #4658 and #4659. While the combined tree was being gated, **#4657 and #4658 were merged to `main` independently** — so the description above overstates what this PR now contains.

The actual diff against current `main` is the perf work alone:

```
src/agent_bom/api/middleware.py         |  14 ++
src/agent_bom/api/routes/demo_estate.py |  61 ++-
src/agent_bom/api/server.py             |   9 +-
src/agent_bom/db/graph_store.py         |  11 +
tests/test_graph_read_path_perf_4654.py | 399 +++++
docs/PRODUCT_METRICS.{json,md}          |   4 +-
```

Sections 1 (correlated findings) and 3 (README) are already on `main` and are **not** part of this diff. Retitled accordingly; #4659 is closed as superseded by this.

The verification still stands and is arguably stronger than it was for the individual branches — the full suite ran on a tree that contained all three sets of changes together: **18951 passed, 191 skipped, 0 failed**, plus `mypy` clean over 831 files, `ruff` clean, and `make preflight` green. That result covers this perf change merged alongside the findings and README work, which is exactly the combination now on `main`.

Unchanged caveat: **UI gates were not run on the merged tree** (`npm ci` hit the disk exhaustion described above). They passed on #4658's own branch, and that work is already merged.

andres-linero (approved):

agent-bom (approved):

PR #4662, opened 2026-08-04
test(perf): calibrate the loop-stall check against the machine

### LINKED ISSUE #4661: ci-regression: CI/CD Pipeline failing on main
Workflow **CI/CD Pipeline** failed on `main`.

- SHA: `d58d114`
- Run: https://github.com/msaad00/agent-bom/actions/runs/30955141694
- Actor: @msaad00

This issue was opened automatically so main failures do not sit unnoticed.

msaad00:
Closes #4661.

## The failure

`test_demo_story_route_does_not_block_the_event_loop`, shipped in #4660, asserted
a fixed ceiling of half the build — 500 ms of a 1,000 ms stand-in. It passed
locally and on the PR, then failed on `main`:

```
AssertionError: event loop stalled 542 ms of a 1000 ms build
  — the estate build is running on the loop
```

**The offload was working.** The heartbeat measures scheduler starvation as well
as loop blocking, and under xdist on a shared CI box the ambient noise alone
reached 542 ms. A fixed wall-clock ceiling cannot distinguish "the route parked
the loop" from "the machine was busy" — so the test was measuring the runner as
much as the code, and it happened to pass on quieter hardware.

That is my defect, introduced with the fix it was meant to protect.

## The fix

Measure an idle noise floor on the same loop moments before the request, then
compare against `max(build * 0.5, noise * 4)`. That asks the question that
actually matters: **is the loop worse off because of the request**, rather than
"is this machine fast".

## Verified not vacuous

A self-calibrating threshold can pass by accident, so the offload was removed to
put the build back on the loop:

```
event loop stalled 2719 ms during a 1000 ms build
  (idle noise floor 1 ms, ceiling 500 ms)
```

Two orders of magnitude clear of the threshold. The check still fails loudly for
the defect it exists to catch, and the production code was restored afterwards
(verified byte-identical).

## Verification

- `pytest tests/test_graph_read_path_perf_4654.py` — 11 passed
- Failure injection above — confirmed still detects an on-loop build
- `ruff check` / `format --check` — clean
- `make preflight` — passed

## Note

`main` is red only on this assertion; the two runs before it were green and the
underlying perf work is sound. This restores `main` without weakening what the
test proves — the ratio floor of `build * 0.5` is unchanged, and the noise term
only ever *raises* the ceiling on a contended machine.

### DIFF
--- a/.claude/worktree-rescue/a05c4f097987b7600.uncommitted.patch
+++ b/.claude/worktree-rescue/a05c4f097987b7600.uncommitted.patch
@@ -1,0 +1,133 @@
+diff --git a/src/agent_bom/output/compliance_export.py b/src/agent_bom/output/compliance_export.py
+index 1cede84d3..f3344d789 100644
+--- a/src/agent_bom/output/compliance_export.py
++++ b/src/agent_bom/output/compliance_export.py
+@@ -156,7 +156,20 @@ def _merged_evidence(report: AIBOMReport, metadata: ComplianceFrameworkMetadata)
+     return merged
+ 
+ 
+-def _bundle_completeness(report: AIBOMReport, mapped_evidence_count: int) -> str:
++def _bundle_completeness(
++    report: AIBOMReport,
++    mapped_evidence_count: int,
++    *,
++    catalog_size: int = 0,
++    controls_with_evidence: int = 0,
++) -> str:
++    """Classify how much of the framework catalog this bundle actually evidences.
++
++    ``complete`` is reserved for a bundle where every catalog control carries
++    evidence. Reporting it whenever *any* evidence existed told an auditor the
++    bundle covered the framework when most controls were ``not_evaluated`` in
++    the very same ZIP.
++    """
+     has_scan_input = any(
+         (
+             report.total_agents,
+@@ -171,6 +184,8 @@ def _bundle_completeness(report: AIBOMReport, mapped_evidence_count: int) -> str
+         return "incomplete"
+     if mapped_evidence_count == 0:
+         return "not_evaluated"
++    if catalog_size and controls_with_evidence < catalog_size:
++        return "partial"
+     return "complete"
+ 
+ 
+@@ -216,7 +231,14 @@ def export_compliance_bundle(
+     # Build policy results
+     mapped_evidence = _merged_evidence(report, metadata)
+     mapped_evidence_count = sum(len(rows) for rows in mapped_evidence.values())
+-    evidence_completeness = _bundle_completeness(report, mapped_evidence_count)
++    evidenced_controls = [ctrl_id for ctrl_id in metadata.catalog if mapped_evidence.get(ctrl_id)]
++    unevaluated_controls = [ctrl_id for ctrl_id in metadata.catalog if not mapped_evidence.get(ctrl_id)]
++    evidence_completeness = _bundle_completeness(
++        report,
++        mapped_evidence_count,
++        catalog_size=len(metadata.catalog),
++        controls_with_evidence=len(evidenced_controls),
++    )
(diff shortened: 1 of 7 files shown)


msaad00:
**Third version, and the approach changed.** Timing was the wrong instrument.

The first version used a fixed 500 ms ceiling and lost to **542 ms of xdist scheduler noise**. The second calibrated against an idle noise floor and lost to a **921 ms stall on a contended runner** with a 1 ms floor. A wall-clock threshold cannot separate *"the route parked the loop"* from *"the machine was busy"* — it was testing the runner, and it would have kept failing intermittently.

It now asserts **thread identity**: the builder must not run on the event-loop thread. That is exact, costs no sleep (the test dropped from ~26s to ~12s), and is the property that actually matters — work on a worker thread cannot block the loop, by construction.

**One thing worth recording, because I nearly shipped it.** An intermediate version compared the builder's thread against the *pytest* thread. It passed — and was **vacuous**. `TestClient` drives the app on its own loop in another thread, so an on-loop build still looks "off-thread". Failure injection caught it: with the offload removed the test still passed. The request is now driven directly on the test's own loop, and the loop thread is captured *inside* the app.

That is the fourth time this session a check reported OK while unable to observe what it claimed to check — and the second one I wrote myself.

Verified non-vacuous. With `anyio.to_thread.run_sync` replaced by a direct call:

```
AssertionError: the estate build ran on the event-loop thread (8394793344)
  — it must be offloaded, otherwise the whole ~1.6s build parks the loop
    for the entire request
```

Production code restored byte-identical afterwards.

Gates: `pytest tests/test_graph_read_path_perf_4654.py` 11 passed, `ruff check` and `format --check` clean, `make preflight` green.

agent-bom (dismissed):

andres-linero (dismissed):

msaad00:
**The 25-minute cancellation was a deadlock, not a slow test — and it was mine.**

The previous version drove the ASGI app with `asyncio.run` to capture the real event-loop thread, while the `story_client` fixture already had the app running on its own loop. The module-level `anyio.CapacityLimiter` binds to whichever loop first uses it, so the second loop deadlocked against it. That is why it burned the full timeout instead of failing.

Fourth version, and the last one that measures anything indirect. The history is worth keeping, because **every failure was the test's fault, not the code's**:

| version | why it failed |
|---|---|
| fixed 500 ms ceiling | lost to 542 ms of xdist scheduler noise |
| noise-calibrated ceiling | lost to a 921 ms stall on a contended runner |
| compare vs pytest thread | **vacuous** — `TestClient` runs the app on its own loop in another thread, so an on-loop build looked off-thread and it passed with the offload removed |
| second loop to capture the real thread | **deadlocked** on the CapacityLimiter, 25 min cancellation |

It now asserts structurally, staying on the fixture's own loop: that `anyio.to_thread.run_sync` was actually **called**, and that the builder ran on a **different thread** than the caller. Work on a worker thread cannot block the loop by construction, and removing the offload removes the call — so there is nothing left to time and no second loop to deadlock.

Verified both directions. With the offload replaced by a direct call:

```
AssertionError: the route never offloaded — anyio.to_thread.run_sync was not
called, so the ~1.6s estate build ran inline on the event loop
```

That failure arrives in **10.8s** rather than hanging. Production code restored byte-identical afterwards.

Gates: 11 passed, `ruff check` + `format --check` clean, `make preflight` green.

Separately — Docker being available locally let me run the two suites that have been failing as "environmental" all session: `test_k8s_live_kind` (2 passed) and `test_online_scan_validation` (4 passed). Both genuinely pass; they were missing a daemon, not broken.

agent-bom (approved):

andres-linero (approved):

PR #4672, opened 2026-08-05
fix(security): one credential predicate across six copies, and close the redaction bypasses

msaad00:
Three security defects from the pre-release audit. The audit said three copies of
the credential predicate — grepping the class found **six**.

## 1. The predicate dropped real credentials from inventory

All eight should be `True`:

```
GOOGLE_APPLICATION_CREDENTIALS -> False    CERTIFICATE -> False
PGPASSWORD                     -> False    OAUTH       -> False
ID_RSA                         -> False    API_KEYS    -> False
SSH_KEYS                       -> False    CREDENTIALS -> False
```

Three narrow additions, **not** a widening — an earlier attempt widened this and
had to be reverted for reporting `CERTIFICATE_PATH` and `DB_CONNECTION_POOL_SIZE`
as credentials:

- **Plurals fold to singular only for words the module already knows**, so
  `API_KEYS`/`CREDENTIALS` match while `DAYS` still matches the configuration word
  `days` rather than becoming an unknown `day`.
- **Compound names** carrying no credential word at all, matched on the *whole*
  name: `pgpassword` (libpq), the ssh-keygen `id_rsa`/`id_dsa`/`id_ecdsa`/
  `id_ed25519` family, `mysql_pwd`.
- **Material words** (`certificate`/`cert`/`oauth`) count alone but are **demoted
  by a locator token** — that demotion is what stops the reverted over-claim
  returning.

Verified both directions on a 50-name sweep (26 real credentials, 24 config):
**0 misses, 0 over-claims**. `AZURE_CLIENT_CERTIFICATE_PATH` and
`SNOWFLAKE_PRIVATE_KEY_PATH` stay credentials; `CERTIFICATE_PATH` does not.

## 2. Six copies, not three

One server with 21 credentials + 18 config vars: the mesh reported **25**, the
graph export **21** — the substring list produced **10 false positives**
(`KEYBOARD_LAYOUT`, `AUTH_MODE`, `AGENT_BOM_TOKEN_ROTATION_DAYS`, …) and **6
misses** (`CA_CERT`, `DATABASE_URL`, `ID_RSA`, …).

Beyond the two named sites, grepping the class found three more of the identical
list — `cli/_analysis.py:81`, `ui/lib/mesh-graph.ts:116`, `ui/app/agents/page.tsx`
— plus a sixth disagreeing copy in a **dead** `except ImportError` fallback in
`graph/builder.py` (`agent_bom.constants` imports only `re`, so the guard could
never fire). All now one answer:

```
credential_names 9 · graph export 9 · mesh stat 9 · CLI "9 cred"
```

## 3. `sanitize_text` leaked a 64-char key in eight shapes

```
bare          API_KEY=a3f9…      json         {"api_key": "a3f9…"}
quoted-double API_KEY="a3f9…"    yaml         api_key: a3f9…
quoted-single API_KEY='a3f9…'    b64-padding  API_KEY=c2Vj…Z2g=
spaced        API_KEY = a3f9…    comma-sep    key=a3f9…,env=prod
```

Two causes: the candidate run excluded quotes, so a quoted value never carried its
key; and `rpartition("=")` split on base64 **padding**, leaving an empty candidate.

**Key side only.** The bare-value entropy fallback is deliberately not imported —
it previously redacted a `hash_ref` and a GuardDuty ARN out of the runtime
taxonomy. Redaction now fires on the key name alone, because a 64-char hex scores
~4.0 bits/char, below the 4.5 threshold: waiting for the *value* to look secret is
exactly what let it through.

**A colon must look like an assignment.** The first version read
`credential_ref:credential_ref:…` as `key: value` and truncated it —
`test_audit_projection_enters_unified_graph` caught it, because the exporter
set-dedups on id, so distinct refs would have collapsed into one phantom node. A
colon now needs a quote (JSON) or trailing whitespace (YAML); `=` needs no proof.

Verified on a 14-line log corpus: content hashes, `sha256:` digests, GuardDuty
ARNs, `trace_id=`, site-packages paths, `exposed_credentials=3` and
`credential_env_vars=OPENAI_API_KEY,GITHUB_TOKEN` all unchanged; only the three
secret-bearing lines redacted. ReDoS check: 200× a 1000-char pathological line =
0.19s.

## Gates

| Gate | Result |
|---|---|
| `pytest` (full) | **19130 passed, 104 skipped**, exit 0 |
| `-k "credential or sanitiz or redact or security or discovery or mesh"` | 1368 passed; widened with `graph or builder`: 2615 passed |
| `ruff check` / `format --check` / `mypy` (6 files) | pass / pass / clean |
| `make preflight`, `check-counts.py` | pass, re-run after merging `main` |
| UI | lint clean · `tsc` clean · vitest **1047 passed** · both builds ok |

## Flagged

- **The UI predicate is a port, not shared code.** Drift is *visible* — both test
  files carry the same fixture names — but nothing mechanically enforces sync.
  Generating it from Python would; that is larger than this fix.
- **`enforcement.check_claude_config` keeps a deliberately separate copy.** Its
  question is "does this hold an inline secret", and the canonical list would make
  it report `CERTIFICATE_PATH=/etc/ssl/ca.pem` as a hardcoded secret. Unchanged,
  as its docstring already documents.
- **Documented limits, asserted not implied:** values under 4 characters are not
  redacted, and a bare unshaped secret with no key still survives.
- `make lint` fails on `main` too — `mypy` dies in installed numpy stubs before
  checking anything. Reproduced with `src/` reverted; environmental.


andres-linero:
"Scan for committed secrets" is red (gitleaks, 1 leak, run 30976982354). Likely source: the new `test_sanitize_env_vars_redacts_every_private_key_label` fixture in `tests/test_security.py` (a synthetic OpenSSH private-key PEM body) and/or `HEX_KEY` in `tests/test_sanitize_text_secret_coverage.py` — both added by this PR to exercise the redaction logic, not real credentials.

`tests/test_security.py` is already in the `.gitleaks.toml` allowlist paths, but that block only covers the `framework_key`/Azurite regexes, not a private-key PEM header, so the new fixture isn't caught by it.

Clearing this needs a scoped `.gitleaksignore` fingerprint or a narrower `.gitleaks.toml` regex (per the existing convention in that file) — leaving that call to a human reviewer rather than widening the secret-scanning allowlist unilaterally. PR left red pending that.

---
_Generated by [Claude Code](https://claude.ai/code)_

msaad00:
Gitleaks fixed — thanks for not widening the allowlist unilaterally, that was the right call.

**The finding was the hex key, not the PEM.** Reproduced locally:

```
RuleID:      generic-api-key
File:        tests/test_sanitize_text_secret_coverage.py
Line:        76
```

That is `HEX_KEY` — the 64-char fixture these redaction tests are built around, chosen precisely because its entropy sits *below* the bare-value threshold so the key name has to be the signal. Non-functional, and the reason the file exists.

**Handled the way `.gitleaks.toml` documents**, rather than by a new regex: a file-scoped `paths` entry, alongside `test_security.py`, `test_dspm_luhn_secrets.py` and the other named secret-DETECTION test files — and mirrored in `.github/secret_scanning.yml` so the two configs do not drift, which that comment explicitly asks for.

**Not `.gitleaksignore`**, deliberately. Its entries are commit-pinned fingerprints (`<sha>:<file>:<rule>:<line>`), so one would go stale the moment this branch is rebased or that line moves — and the file's own comment scopes it to *history-only* false positives. This fixture is current, not historical.

The path regex matches exactly one filename, so scoping is structural: a real secret anywhere else under `tests/` is still flagged. Verified the finding is cleared — 0 findings in that file.

andres-linero (approved):

agent-bom (approved):

msaad00:
**Full-suite evidence on the combined tree.**

Before #4673 merged independently, both branches were merged together and gated as one. The point of combining is exactly this: two branches green in isolation can regress together, and neither ran the other's tests.

```
pytest -m "not network" -p no:randomly
19069 passed, 190 skipped, 16 deselected — 0 failed  (21m42s)
```

Plus, on that same merged tree: `make preflight` green, `ruff check src/ tests/ scripts/` clean, `mypy` clean over 832 files, zero conflict markers.

So this change is verified against the read-path work that is now on `main`, not only in isolation.

**One thing that surfaced from linting the merged tree**, worth noting beyond this PR: `ruff check` flagged an F841 dead variable in `scripts/generate_doc_architecture_svgs.py` — mine, left over from the two-row persona layout. It is invisible to CI because **CI lints only `src/`**, so `scripts/` and `tests/` drift unchecked. Fixed in that combine branch and worth widening the lint scope separately; the release tooling lives in `scripts/`.

Deterministic order (`-p no:randomly`) was used deliberately — after the `google.cloud` namespace fix, a random-order failure would report on ordering rather than on this change.

PR #4678, opened 2026-08-05
chore(toolchain): pin one ruff, gate formatting, and reformat

### LINKED ISSUE #4676: Reconcile the three ruff versions gating this repo
Three different ruff versions currently gate the same tree:

| source | version |
|---|---|
| `uv.lock` (what CI resolves) | **0.16.1** |
| `.pre-commit-config.yaml` | **0.9.7** |
| `pyproject.toml` dev extra | `ruff>=0.4` |

The lint rules agree across these, so `ruff check` is not currently affected —
but the **formatter** style changed between 0.9 and 0.16. Running
`ruff format --check src/ tests/` with 0.16.1 reports ~240 files as unformatted,
because the tree was formatted by the pre-commit hook's 0.9.7.

Consequences today:

- A contributor whose local ruff matches `uv.lock` cannot run `make format`
  without producing a 240-file diff.
- There is no `ruff format --check` gate anywhere, so the divergence is
  invisible until someone runs the formatter.
- The pre-commit hook and CI could begin disagreeing about `ruff check` results
  on any future rule change, with no gate to catch it.

Proposed fix, as a standalone change (deliberately not bundled with release
work, since the reformat commit is large and mechanical):

1. Pin `.pre-commit-config.yaml` to the same ruff version as `uv.lock`.
2. Run `ruff format` once over the linted scope and commit the result as a
   single formatting-only commit, so `git blame` can skip it via
   `.git-blame-ignore-revs`.
3. Add `ruff format --check` to the Makefile and to CI's lint job, using the
   same `LINT_PATHS` variable that #4675 introduces.
4. Tighten the `pyproject.toml` floor so `>=0.4` cannot resolve a formatter
   generations behind the lock.

Found while widening the lint scope in #4675.

msaad00:
## Summary

Three sources pinned ruff independently, and they disagreed:

| source | version | what it gates |
|---|---|---|
| `uv.lock` | **0.16.1** | CI |
| `.pre-commit-config.yaml` | **0.9.7** | a contributor's commit hook |
| `pyproject.toml` dev extra | `ruff>=0.4` | whatever pip resolves |

The lint *rules* agree across that range, so nothing ever went red. The
*formatter* does not: the committed tree was 0.9.7's output while CI resolved a
version that rewrites 249 files. Nothing compared the two, because there was no
`ruff format` gate anywhere — `make format` was a 249-file diff waiting for
whoever ran it first.

Closes #4676.

## Commits, deliberately separate

1. **`chore(toolchain)`** — the pins agree, the dev-extra floor rises to
   `>=0.16.1`, `make format-check` is added and wired into CI's lint job over
   the `LINT_PATHS` from #4675, plus `tests/test_toolchain_pin_agreement.py`.
2. **`style`** — the reformat. 249 files, mechanical, no behaviour change.
3. **`chore`** — `.git-blame-ignore-revs`, see the note below.
4. **`chore(deps)`** — relock for the new floor.

Reviewing commits 1, 3 and 4 is the whole review; commit 2 is `ruff format`
output.

## The blame-ignore file carries no SHA, on purpose

I first recorded the reformat commit's SHA. Rebasing onto `main` changed it, and
this repository **squash-merges** — so the SHA that finally holds the reformat
does not exist until this PR lands. Any SHA written here is dead on arrival, and
a dead entry is invisible: git skips nothing and reports nothing.

So the file ships with the convention and the follow-up instruction, and the
entry goes in as a one-line commit on `main` afterwards. The test asserts what
actually breaks silently — an abbreviation (git errors) or a SHA that does not
resolve (git ignores it quietly) — and permits an empty list, which is the only
honest state before the merge.

## Verification

- All four tests confirmed **red** first: `v0.9.7 != v0.16.1`,
  `ruff>=0.4 != ruff>=0.16.1`, no `format-check:` target, no blame-ignore file.
- **Full suite on the reformatted tree: 19,138 passed, 117 skipped, 0 failed**
  (`uv run --frozen --extra dev-all pytest -n auto tests/`).
- `make format-check` — 2064 files already formatted. `make lint-ruff` — all
  checks passed. Both through `uv run --frozen`, i.e. the locked 0.16.1 rather
  than a locally installed ruff.
- `mypy src/` — no issues in 832 source files. `make preflight` — passed.
- `uv lock --check` — clean. The relock also tightened one numpy resolution
  marker; that is uv 0.10.9's own output for the current manifest, not a
  dependency change.

## Note

An earlier full-suite run showed `tests/test_audit_otlp_logs.py` failing. That
was my invocation missing the optional `otel` extra, not the reformat — with
`--extra otel` it is 10 passed. Separately worth knowing: that test hard-fails
rather than skipping when an optional extra is absent. Not changed here.


msaad00:
Rebased onto `main` now that #4677 has merged.

That overlap was worth checking rather than assuming: #4677 edits
`scripts/check_surface_freshness.py` and `tests/test_surface_freshness.py`, the
same two files the reformat commit rewrites. Two green branches can regress when
combined, so the gates were re-run **on the merged result**, not on either side:

- `make format-check` — 2064 files already formatted, i.e. #4677's new lines are
  0.16.1-clean under the reformat
- `make lint-ruff` — all checks passed
- `make preflight` — passed
- **full suite: 19,142 passed, 117 skipped, 0 failed**
  (`uv run --frozen --extra dev-all pytest -n auto tests/`)

No behaviour change from the rebase; the diff against `main` is unchanged apart
from those two files now carrying #4677's content.

agent-bom (approved):

andres-linero (approved):

PR #4680, opened 2026-08-05
fix(findings): stop an unknown source label from erasing a known finding type

msaad00:
Found by running the G5–G8 pre-release scale gates against real Postgres at 1M findings — but it is not a scale or backend bug. **It reproduces on SQLite with ten findings in 1.4 seconds.**

## The defect

`POST /v1/findings/bulk` is the connector ingest path. Its `source` is a free-text provenance label whose default is literally `"api"`. `domain_for_row()` and `lenses_for_row()` parsed that same field as the `FindingSource` enum — and parsed source and type in **one `try` block**, so an unrecognized label threw the finding type away with it.

A row plainly typed `CVE` therefore derived no domain and no lenses:

```
domain_for_row({"source": "api", "finding_type": "CVE"})  ->  None
lenses_for_row({"source": "api", "finding_type": "CVE"})  ->  frozenset()
```

So **every** `?domain=` query returned nothing for connector-ingested findings, while the identical finding discovered by a scan filtered correctly. Verified at 1M findings: `/v1/findings` returned `total: 1000001`, and `/v1/findings?domain=vuln` returned `0`. The overview coverage lane's own `href` (`/findings?domain=vuln`) lands on an empty page. Because `"api"` is the default, the out-of-the-box path hit this 100% of the time.

No data was lost — `/v1/findings` reconstructs the full row from the ledger. Only the derivation failed.

## The fix

Source and type are parsed independently, and the precedence chain accepts an unknown source. The type-decisive rules never needed it: a CVE is vulnerability management wherever it came from. Only the source-routed fallbacks are skipped.

Two things it deliberately does **not** do:

- **It does not invent a lane.** With neither a parseable type nor source it still returns `None`, so the caller picks the default rather than filing arbitrary findings under AISPM.
- **It does not add a second copy of the mapping.** The source-optional function *is* the implementation and the public one delegates, so the row-level and object-level entry points cannot drift — that is the defect class that put four agent-tool detectors and five SARIF severity tables in this tree.

## Verification

| | |
|---|---|
| bulk CVE | → `vuln` |
| bulk `CREDENTIAL_EXPOSURE` | → `aspm` |
| bulk `CIEM_OVER_PRIVILEGE` | → `cspm` |
| bulk `SENSITIVE_DATA` | → `dspm` |
| unknown type | → `None` (unchanged, honest) |
| `SBOM` + `CVE` | → `{vuln, aspm}` (unchanged) |

Asserted **end-to-end through `/v1/findings`**, not just on the mapping: the scope filter runs on enriched rows inside the store, so a unit test on the helper could pass while the route still returned an empty page.

- `test_finding_scope.py` + `test_bulk_ingest_domain_filter.py`: 30 passed
- facet pushdown / export contract / surface parity / hub bulk ingest / lifecycle: 68 passed, 3 skipped
- `ruff format`, `ruff check`, `mypy` (832 files), `make preflight`: clean

## Also in this pass — two release gates that under-reported

- **`preflight_release.sh` printed "Pre-flight OK — safe to tag" when run without a version**, because that argument is what enables the bump check. Every managed reference could be a release behind and the verdict still read green. The version is now required (exit 2 with a usage message). This is the same "a gate that cannot see what it gates" shape the last sweep closed elsewhere — in the release script itself.
- **`bump-version.py --check` printed "Updated N occurrence(s)"** for a run that writes nothing (confirmed: the tree stays clean). During a release that reads as "the bump is done." Check mode now says `Would update` / `DRIFT`.

And the 0.99.0 changelog now records #4677, which was merged after the entry was written.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,13 @@ some numbers you may already be reporting will move.
   pagination link would have carried the credential off-origin. It now refuses
   any pagination URL that leaves `hub.docker.com`, bounds the loop, and fails
   rather than planning deletions against a partial tag list.
+- The GHCR freshness probe had the same off-origin pagination weakness and is
+  now pinned to `ghcr.io` on the same terms. In the same pass, the surface
+  freshness monitor's tool-count assertion silently switched itself off when
+  `--expected-tool-count` was unset: only the Glama probe derived a fallback, so
+  a bare `python scripts/check_surface_freshness.py` reported a listing fresh
+  without ever checking the count. CI passed the flag, so only humans saw the
+  weaker check.
 - Configuration was being counted as credentials (`CERTIFICATE_PATH`,
   `DB_CONNECTION_POOL_SIZE`), and real credentials were being missed
   (`PGPASSWORD`, `ID_RSA`, `GOOGLE_APPLICATION_CREDENTIALS`, and every plural
--- a/docs/PRODUCT_METRICS.json
+++ b/docs/PRODUCT_METRICS.json
@@ -28,7 +28,7 @@
     },
     {
       "name": "Test files",
-      "value": 1136,
+      "value": 1137,
       "source": "tests/",
       "notes": "Counts files matching test_*.py."
     },
--- a/docs/PRODUCT_METRICS.json
+++ b/docs/PRODUCT_METRICS.json
@@ -46,7 +46,7 @@
     },
     {
       "name": "Python modules",
-      "value": 832,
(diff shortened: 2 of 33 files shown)


agent-bom (approved):

andres-linero (approved):

msaad00:
Superseded by #4682, which combines this with the other pre-release correctness fixes into one reviewable change (plus a third gate gap found while bumping to 0.99.0). Branch retained.

PR #4681, opened 2026-08-05
fix(ui): hash the beforeInteractive scripts the browser actually executes

msaad00:
**The dashboard's CSP blocks an inline script on every page, in production.** Confirmed against the live released demo (0.98.3), not just a local build — so this ships today.

## Why the allowlist missed it

`next/script` with `strategy="beforeInteractive"` does **not** emit an inline `<script>` tag. Next serializes the body into `self.__next_s`, and its `appBootstrap` runtime creates a script element whose text is exactly the `children` string, then appends it — so the browser evaluates that text and checks it against `script-src-elem`.

`generate_ui_csp_hashes.py` scans the exported HTML for `<script>` tags. It hashed the **wrapper push** and never the payload that actually runs:

```
theme-bootstrap runtime text hash: sha256-CVIz+8cEN0ddhFEe/IMBU5uIChZx5ZAwQQKl0dHHVuQ=
in CSP allowlist? False
```

Captured directly from a `securitypolicyviolation` listener:

```json
{ "directive": "script-src-elem", "blockedURI": "inline",
  "sourceFile": ".../chunks/3794-321373812a1a8bf8.js", "lineNumber": 32 }
```

## Impact

The blocked script is the theme bootstrap — it reads the stored theme and applies it before first paint. Blocked, that means **a flash of the wrong theme on every page load** for anyone whose stored preference isn't the default, plus a CSP error in the console on every page of a security product's own dashboard.

## Why no existing gate caught it

The endpoint returns 200, the page renders, and the HTML export is byte-for-byte correct. The defect exists only in what the browser *does* with it. It took the pre-release rendered-UI pass — the gate that exists precisely because two of the four 0.96.0 misses were invisible to API checks and obvious on the page.

It is also the recurring one-judgement-two-implementations shape: `ui/lib/security-headers.mjs` derives the theme hash from `THEME_BOOTSTRAP_SCRIPT` **correctly**, while the shipped FastAPI server reads `csp-hashes.json`, which was blind to it. Two CSP sources; the one that ships was the wrong one.

## The fix

The generator now parses `self.__next_s` entries and hashes any inline `children` body — covering future `beforeInteractive` scripts rather than special-casing the theme one.

## Verification

- 118 → 119 script hashes.
- Playwright `securitypolicyviolation` listener: **1 violation per page → 0**.
- Full rendered pass across 16 pages against a seeded estate: **8/16 clean → 14/16**. The two remaining are correct behavior, not defects — `/activity` returns a clear 400 (`SNOWFLAKE_ACCOUNT env var not set. Activity requires Snowflake.`) and `/gateway` 403s because anonymous resolves to `viewer`, which lacks `audit_read`.
- `tests/test_ui_csp_hashes.py` gains a regression test asserting the beforeInteractive body is in the manifest, plus one pinning that external-`src` entries contribute no inline hash. Dashboard suites: 15 passed, 1 skipped.

`src/agent_bom/ui_dist/` is gitignored and rebuilt by the release job, so no generated artifact is committed here — the fix is in the generator the release runs.

### DIFF
--- a/scripts/generate_ui_csp_hashes.py
+++ b/scripts/generate_ui_csp_hashes.py
@@ -7,6 +7,7 @@ import argparse
 import base64
 import hashlib
 import json
+import re
 from html.parser import HTMLParser
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


msaad00:
Superseded by #4682, which combines this with the other pre-release correctness fixes into one reviewable change (plus a third gate gap found while bumping to 0.99.0). Branch retained.

PR #4689, opened 2026-08-06
fix(demo): make the AI BOM read the estate instead of reporting zero

msaad00:
The AI BOM page rendered **0 agents / 0 MCP servers / 0 models / 0 packages / 0 findings** on an estate holding **48 agents, 25 MCP servers and 97 tools**.

Its own heading promises *"live inventory from connected agents, cloud accounts, endpoints, and runtime — not a static upload."* It was showing nothing.

## Root cause

`/v1/agent-bom/manifest` is built from exactly two stores:

```python
fleet_agents = _get_fleet_store().list_by_tenant(tenant_id)
observations = _get_mcp_observation_store().list_by_tenant(tenant_id)
return build_control_plane_agent_manifest(fleet_agents, observations, ...)
```

The demo bootstrap seeds the **graph**, the **findings** and the **identities** — and never touched either of those. So every count on the page was structurally zero regardless of estate size.

## The fix

`seed_showcase_fleet_and_runtime` follows the same contract as the existing `seed_showcase_identities`: idempotent, isolated behind its own `try/except` so a failure can't block the rest of the bootstrap, and only under demo-estate mode.

```
```

**Variety is deliberate, not decoration.** Fleet lifecycle spreads to **28 approved · 10 pending review · 5 quarantined · 5 discovered**. Every agent sharing one state is the tell that nothing is actually being governed — the exact impression the estate exists to avoid. Servers likewise vary across runtime-observed and gateway-registered.

Rows carry operator context — real names (`member-copilot-00`, `Care Gap Analytics MCP`), owner, environment, trust score, transport, auth mode — because an inventory of bare identifiers can't be reasoned about.

## Verification

Asserted through `build_control_plane_agent_manifest` — the function the route actually calls — **not** on the seeder's return value, because "48 seeded" was already true while the page showed zero.

- `test_demo_estate_ai_bom_manifest.py` — 4 tests: manifest is non-empty, lifecycle has ≥3 states with `approved` dominant, rows carry name/owner/environment/transport/auth, and re-seeding is a no-op (the bootstrap runs on every start).
- demo estate + showcase seeding suites: **50 passed**
- `ruff` · `mypy` clean

## Known gap deliberately NOT fixed here

`summary.tools` is still **0**. `_observed_server()` emits no `tools` key and `MCPObservation` carries no tools field — so the tool count is structurally zero for **any** control-plane tenant, not just the demo. That's a store schema change and doesn't belong in this PR, but it means the AI BOM still under-reports what the servers expose.

### DIFF
--- a/src/agent_bom/demo_estate/bootstrap.py
+++ b/src/agent_bom/demo_estate/bootstrap.py
@@ -12,6 +12,7 @@ from agent_bom.api.pipeline import _now
 from agent_bom.api.store import DEMO_ESTATE_TRIGGERED_BY
 from agent_bom.demo_estate.showcase_graph import (
     SHOWCASE_TENANT,
+    seed_showcase_fleet_and_runtime,
     seed_showcase_graph_if_empty,
     seed_showcase_identities,
 )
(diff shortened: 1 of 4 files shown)


agent-bom (approved):

andres-linero (approved):

msaad00:
Superseded by #4691, which combines this with the other estate-readability fixes into one reviewable change. Branch retained.

PR #4690, opened 2026-08-06
feat(demo): name estate resources after workloads, not coordinates

msaad00:
Resources were named `f"{provider}-{service}-{account:02d}-{slot:03d}"`:

```
aws-iam-03-020     snowflake-data-00-004     azure-storage-05-041
```

That's a position in a loop, not a name. **Nothing in the estate could be reasoned about** — an over-privileged role called `aws-iam-03-020` tells an operator nothing about blast radius, and 3,860 of them read as *generated* rather than *discovered*, which is the one impression a demo estate has to avoid.

## After

| Type | Before | After |
|---|---|---|
| `iam_role` | `aws-iam-03-020` | `prior-auth-reviewer-development`, `clinical-analytics-reader-development` |
| `bucket` | `aws-s3-01-007` | `member-exports-prod`, `provider-rosters-staging` |
| `table` | `snowflake-data-00-004` | `care-gaps-staging`, `encounters-staging` |
| `cluster` | `gcp-gke-02-013` | `member-ai-prod`, `population-health-development` |
| `function` | `aws-lambda-04-022` | `attachment-virus-scan-staging`, `eob-renderer-staging` |

Names come from workload pools a payer/provider AI platform actually runs, so a least-privilege failure on `phi-export-approver-prod` is a sentence a reviewer can act on.

## Details that matter

- **Pools are keyed by resource class, not provider** — the same business capability appears across clouds, the way it does in a real multi-cloud estate.
- **The environment suffix carries information.** `claims-etl-prod` and `claims-etl-staging` are genuinely different blast radii.
- **The numeric tail only appears when a pool wraps inside one account**, so the common case reads as a name rather than a serial number.
- **Cluster pool entries carry no environment token of their own** — that was producing `member-ai-prod-development`.
- **Deterministic** by `(provider, account, type, slot)`, so snapshots, the drift lens and pinned tests stay stable across builds.

## Gated on the property, not a name list

- No asset may carry the coordinate shape — a regex, so adding workloads can't quietly reintroduce it
- **Ids stay unique** — a wrapping pool must not collapse two resources onto one asset id
- Names are **stable between builds** — the drift lens compares by id, so churn would read as change
- **The hero incident chain keeps its ids** (`member-copilot-prod`, `member-ai-prod/ai-prod/member-copilot`, `nh_prod/analytics/phi/patient_summary`) — the graph, correlations and demo story all reference those

**132 passed** across the eight estate/demo suites · `ruff` · `mypy` clean.

### DIFF
--- a/src/agent_bom/demo_estate/enterprise_scale.py
+++ b/src/agent_bom/demo_estate/enterprise_scale.py
@@ -100,6 +100,111 @@ _RESOURCE_MIX: dict[str, tuple[tuple[str, str], ...]] = {
     ),
 }
 
+# What the workloads are FOR. Resource names were previously
+# ``f"{provider}-{service}-{account:02d}-{slot:03d}"`` -> ``aws-iam-03-020``,
+# which is a coordinate, not a name. Nothing in the estate could be reasoned
+# about: an over-privileged role called ``aws-iam-03-020`` tells an operator
+# nothing about blast radius, and a whole inventory of them reads as generated
+# rather than discovered — which is exactly the impression a demo must avoid.
+#
+# These are the workloads a payer/provider AI platform actually runs, so a role
+# named ``prior-auth-reviewer`` failing least-privilege is a sentence a reviewer
+# can act on. Kept as one pool per resource CLASS rather than per provider, so
+# the same business capability shows up across clouds the way it does in a real
+# multi-cloud estate.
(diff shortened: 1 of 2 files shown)


msaad00:
Superseded by #4691, which combines this with the other estate-readability fixes into one reviewable change. Branch retained.

PR #4691, opened 2026-08-06
fix(demo): make the estate readable — AI BOM inventory, workload names, and a unit on the coverage lanes

msaad00:
Supersedes #4689 and #4690, plus one more fix found while reviewing the overview. All three are the same defect class: **the data existed and the page didn't say so.**

---

## 1. The AI BOM read zero on a populated estate

It rendered **0 agents / 0 MCP servers / 0 models / 0 packages** while the estate held 48 agents, 25 MCP servers and 97 tools. Its own heading promises *"live inventory from connected agents … not a static upload."*

`/v1/agent-bom/manifest` is built from exactly two stores — the fleet registry and the MCP observation store — and the demo bootstrap seeds the graph, findings and identities, but **neither of those**. Every count was structurally zero regardless of estate size.

```
```

`seed_showcase_fleet_and_runtime` follows the existing `seed_showcase_identities` contract: idempotent, isolated behind its own `try/except`, demo-mode only.

**Fleet lifecycle spreads to 28 approved · 10 pending review · 5 quarantined · 5 discovered.** Every agent in one state is the tell that nothing is being governed.

Asserted through `build_control_plane_agent_manifest` — the function the route calls — **not** the seeder's return value, because "48 seeded" was already true while the page showed zero.

---

## 2. Resources were named after their loop position

```
aws-iam-03-020     snowflake-data-00-004     azure-storage-05-041
```

A coordinate is not a name. An over-privileged role called `aws-iam-03-020` tells an operator nothing about blast radius, and 3,860 of them read as *generated* rather than *discovered*.

| Type | After |
|---|---|
| `iam_role` | `prior-auth-reviewer-development`, `phi-export-approver-prod` |
| `bucket` | `member-exports-prod`, `provider-rosters-staging` |
| `table` | `care-gaps-staging`, `encounters-staging` |
| `cluster` | `member-ai-prod`, `population-health-development` |
| `function` | `attachment-virus-scan-staging` |

**Zero coordinate-named assets remain.** Deterministic by `(provider, account, type, slot)` so snapshots and the drift lens stay stable. Gated on the property — no asset may match the coordinate regex, ids stay unique (a wrapping pool must not collapse two resources onto one id), names are stable between builds, and the hero incident chain keeps its ids.

---

## 3. The coverage lanes showed a magnitude with no unit

`CSPM 1610`, `DSPM 107` — under a heading like CSPM, a bare number reads as assets, accounts, VMs or data stores depending on the reader. **All wrong.** They are *findings* in that posture lane, which is exactly what the severity chips beneath them sum to (`300 + 633 + 358 + 258 + 61 = 1610`).

So `DSPM 107` meant 107 findings in the data-security lane — not 107 connected data stores, which is the natural misreading.

The number now carries `findings`, and the caption says *"Open findings per posture discipline — not assets or accounts"*, with the non-additive warning kept.

---

## Verification — on the combined tree, not the branches separately

- Estate + AI BOM + correlation suites: **73 passed**
- Bootstrap, showcase seeding, demo surfaces, graph projection: **66 passed**
- **Full UI suite: 168 files, 1,048 tests passed**
- `ruff format` · `ruff check` · `mypy` (835 files) · `make preflight` — all clean

### DIFF
--- a/docs/PRODUCT_METRICS.json
+++ b/docs/PRODUCT_METRICS.json
@@ -1,5 +1,5 @@
 {
-  "generated_on": "2026-08-05",
+  "generated_on": "2026-08-06",
   "version": "0.99.0",
   "metrics": [
     {
@@ -28,7 +28,7 @@
     },
     {
       "name": "Test files",
-      "value": 1139,
+      "value": 1140,
       "source": "tests/",
       "notes": "Counts files matching test_*.py."
     },
--- a/docs/PRODUCT_METRICS.md
+++ b/docs/PRODUCT_METRICS.md
@@ -5,7 +5,7 @@
 This appendix is the canonical home for volatile product counts.
 Keep counts out of public positioning copy and update this file from the repo instead of hand-editing numbers.
 
-- Generated on: `2026-08-05`
+- Generated on: `2026-08-06`
 - Version: `0.99.0`
 
 | Metric | Value | Source | Notes |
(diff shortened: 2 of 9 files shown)


agent-bom (approved):

andres-linero (approved):

msaad00:
Superseded by #4700, which combines the main-unblocking test fix with the estate readability changes into one reviewable PR.

PR #4700, opened 2026-08-06
fix: unblock main, and make the estate readable

msaad00:
Supersedes #4691 and #4692. **Merge this first — it unblocks `main`, which is currently red.**

---

## 1. `main` is red, and it's my regression

#4687 split `AUTH_REQUIRED` / `AUTH_TIMEOUT` out of Smithery's fatal terminal statuses. `test_deployment.py` pinned the old combined literal `FAILURE_SCAN|AUTH_REQUIRED|AUTH_TIMEOUT`, which stopped existing.

**My miss:** I ran `test_ci_workflow_contract` and `test_release_storefront_contract` when making that change and never ran `test_deployment.py` — two suites assert against the same workflow file and I only ran the ones whose names matched what I was editing. Verified by reproducing on clean `main`.

Rewritten to assert behaviour rather than a literal, including a **negative** assertion that `AUTH_REQUIRED` is never regrouped with the fatal statuses — folding it back in silently returns to failing every release for a publish that worked.

This is also why several open dependabot PRs show red: #4696 fails on this same test.

---

## 2. The AI BOM read zero on a populated estate

`0 agents / 0 MCP servers` while the estate held 48 agents and 25 servers. The manifest is built from the fleet registry and MCP observation store — and the demo bootstrap seeds the graph, findings and identities, but **neither of those**.

```
```

Fleet lifecycle spreads to **28 approved · 10 pending review · 5 quarantined · 5 discovered** — every agent in one state is the tell that nothing is being governed.

Asserted through `build_control_plane_agent_manifest`, the function the route calls, **not** the seeder's return value: "48 seeded" was already true while the page showed zero.

---

## 3. Resources were named after their loop position

`aws-iam-03-020`, `snowflake-data-00-004`. A coordinate is not a name — an over-privileged role called `aws-iam-03-020` tells an operator nothing about blast radius, and 3,860 of them read as generated rather than discovered.

Now `phi-export-approver-prod`, `member-exports-prod`, `care-gaps-staging`, `member-ai-prod`. **Zero coordinate-named assets remain.**

Gated on the property: no asset matches the coordinate regex, ids stay unique (a wrapping pool must not collapse two resources onto one id), names are stable between builds, and the hero incident chain keeps its ids.

---

## 4. The coverage lanes showed a magnitude with no unit

`CSPM 1610`, `DSPM 107` — under a three-letter heading, a bare number reads as assets, accounts, VMs or data stores. All wrong. They are **findings**, which is what the severity chips sum to (`300+633+358+258+61 = 1610`).

`DSPM 107` meant 107 findings in the data lane, **not** 107 connected data stores — the natural misreading. Now labelled, with the caption stating "not assets or accounts" and keeping the non-additive warning.

---

## Verification — combined tree

- deployment / AI BOM / estate scale: **84 passed**
- estate + correlation suites: **73 passed** · bootstrap, seeding, surfaces, projection: **66 passed**
- **Full UI suite: 168 files, 1,048 tests**
- `ruff` · `mypy` (835 files) · `make preflight` — clean

### DIFF
--- a/docs/PRODUCT_METRICS.json
+++ b/docs/PRODUCT_METRICS.json
@@ -1,5 +1,5 @@
 {
-  "generated_on": "2026-08-05",
+  "generated_on": "2026-08-06",
   "version": "0.99.0",
   "metrics": [
     {
@@ -28,7 +28,7 @@
     },
     {
       "name": "Test files",
-      "value": 1139,
+      "value": 1140,
       "source": "tests/",
       "notes": "Counts files matching test_*.py."
     },
--- a/docs/PRODUCT_METRICS.md
+++ b/docs/PRODUCT_METRICS.md
@@ -5,7 +5,7 @@
 This appendix is the canonical home for volatile product counts.
 Keep counts out of public positioning copy and update this file from the repo instead of hand-editing numbers.
 
-- Generated on: `2026-08-05`
+- Generated on: `2026-08-06`
 - Version: `0.99.0`
 
 | Metric | Value | Source | Notes |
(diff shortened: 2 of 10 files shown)


agent-bom (approved):

andres-linero (approved):

msaad00:

PR #4701, opened 2026-08-06
chore(deps): bump github/codeql-action to v4.37.5 across every workflow

msaad00:
Consolidates **#4695, #4696, #4697 and #4698** — four dependabot PRs bumping four sub-actions of the *same* action to the *same* version.

Splitting them meant four CI runs and four reviews for one upgrade, plus a window where the pinned SHAs disagreed with each other across workflows depending on merge order.

## What moves

`init`, `autobuild`, `analyze` and `upload-sarif` — every reference, in all **five** workflow files that pin them (`codeql`, `release`, `container-rescan`, `pr-security-gate`, `post-merge-self-scan`).

```
f205ea1c3313d32999d8d6a48b4f6530d4437b38  # v4.37.4
→ d1ba80a13dd99fba24a470575428917156a28b43  # v4.37.5
```

The old SHA no longer appears anywhere in `.github/workflows/`. The new SHA is taken from dependabot's own diff rather than resolved by hand, so the pin is exactly what it verified.

`actionlint` clean on every touched workflow.

## Note on CI

This branch is cut from `main`, which is currently red — `test_deployment.py::test_smithery_publish_is_gated_on_server_card_and_catalog_parity` fails here for that reason, not because of this change. **#4700 fixes it**; merge that first and this goes green. Everything else passes (255 passed, 6 skipped across the workflow/deployment/CI suites).

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -33,16 +33,16 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@f205ea1c3313d32999d8d6a48b4f6530d4437b38 # v4.37.4
+        uses: github/codeql-action/init@d1ba80a13dd99fba24a470575428917156a28b43 # v4.37.5
         with:
           languages: python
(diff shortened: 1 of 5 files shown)


msaad00:

PR #4702, opened 2026-08-06
chore(deps): take @dagrejs/dagre 3.1.0 and restore bundle headroom

msaad00:
Supersedes #4694, which failed `UI Validate`:

```
FAIL total client JS: 3654.7 KiB / budget 3648.0 KiB
```

**6.7 KiB over.** dagre is the layout engine behind the graph and DAG views, so staying current on it is worth the space — but the overrun is the **library's**, not a new route of ours. That's worth recording, because every prior raise in this file was our own feature earning the space.

## The raise

`3648 → 3664 KiB`, on exactly the terms of every raise above it in the file: ~16 KiB of bounded headroom so routine bundler variance stops failing the gate. **Largest-chunk and shared-runtime budgets are unchanged**, so this cannot mask a regression in either.

```
PASS total client JS:      3654.7 KiB / budget 3664.0 KiB
PASS largest chunk:         379.8 KiB / budget  927.7 KiB
PASS shared app runtime:    416.9 KiB / budget  439.5 KiB
```

Measured locally at **3654.7 KiB — the same number Linux CI produced** — so the remaining 9.3 KiB of headroom is real, not platform variance.

Full UI suite: **168 files, 1,048 tests passed**.

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -8,7 +8,7 @@
       "name": "ui",
       "version": "0.99.0",
       "dependencies": {
-        "@dagrejs/dagre": "^3.0.0",
+        "@dagrejs/dagre": "^3.1.0",
         "@tanstack/react-table": "^8.21.3",
         "@tanstack/react-virtual": "^3.14.9",
         "@xyflow/react": "^12.11.2",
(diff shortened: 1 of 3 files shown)


msaad00:

PR #4725, opened 2026-08-07
fix(ui): reconcile truncation counts, thin the band tower, and make drawers readable

msaad00:
Three defects from the UI walkthrough.

## 1. The truncation banner printed impossible arithmetic

```
Showing 5,264 of 4,904 · 4,798 omitted
```

More shown than exist, and no two figures reconciling. The component read `returned` and `total` from the API envelope but `omitted` from a local prop — **three numbers, two derivations, one sentence**:

```ts
const returned = completeness?.returned ?? visibleCount;   // envelope
const total    = completeness?.total ?? ...;               // envelope
const omitted  = omittedCount ?? ...;                      // local prop
```

Any caller passing both hit it every time. The envelope now wins **wholesale** when it carries counts; local props are the fallback for callers with no envelope, never a supplement to one. A total below what was returned is upstream incoherence, so it widens to what we can stand behind rather than printing an impossibility.

## 2. Eight control bands above the content

The investigation page stacked title → lens row → investigation loop → deploy gate → exposure paths → snapshot → large-estate notice → metric tiles, pushing the ranked paths and the graph toward the fold.

- The **investigation loop** is a control, so it rides in the toolbar row instead of owning a band with an explanatory paragraph.
- **"Should I deploy?"** is a deliberate, occasional action — now disclosure-gated rather than costing a permanent band on every visit.

Controls recede; content is the hero.

## 3. Drawers were a mile-long column of one-word cards

The runtime-alert drawer rendered every field as its own full-width bordered card: **nine boxes, one short value each**, ~90px of vertical per word — and five of them repeated facts already in the header subtitle. With event/request/trace IDs it reaches twelve.

Fields now lay out as a **two-column label/value grid** (twelve values ≈ six rows), and width is **reader-controlled and persisted** via a shared `useDrawerWidth` hook so the next drawer inherits it. The handle is focusable and responds to arrow keys — a pointer-only resize is unusable for anyone who can't drag.

Storage is guarded on **both** read and write. Storage is absent or throws in private-browsing and blocked-cookie contexts, and the unguarded read broke an unrelated proxy-dashboard test — exactly the failure a user in private browsing would have hit.

## Verification

- **7 new tests**: 3 vitest pinning the count invariants (never shows more than the total, shown + omitted == total, no source blending), 4 vitest on the drawer (grid layout, resize control, persistence, no information lost), 2 E2E on layout (ranked paths within the first viewport at 1440×900, deploy gate collapsed by default)
- `make preflight` · `tsc` · `eslint` · `next build` clean
- **35/35 E2E · 1055 vitest**
- Bundle on a clean `npm ci`: **3660.0 / 3664.0 KiB** — all three budgets pass without a raise

One process note: my first bundle measurement read 3679.2 KiB and looked like a regression. It was a stale `node_modules` carrying Next 16.3.0 from a different branch; `npm ci` against this branch's lockfile gives the real number.


msaad00:
Combined into #4728 with the root-cause fix for the metrics-snapshot churn that was blocking both.

PR #4726, opened 2026-08-07
fix(ui): make the gates fail loudly when they are measuring the wrong thing

msaad00:
Two gates could silently report a number describing something other than this branch. **Both did, in the same session.**

## 1. `bundle:check` measured the wrong dependency set

Switching between branches with different lockfiles leaves the previous branch's packages installed. The budget then describes packages this branch doesn't use.

```
measured:  3679.2 KiB / budget 3664.0 KiB   FAIL   ← node_modules had a newer Next
after npm ci: 3660.0 KiB / budget 3664.0 KiB PASS   ← the branch's actual number
```

That failure looked exactly like a regression the branch had introduced. **The obvious response would have been to raise the budget** — permanently loosening a real guard on the strength of a bad measurement, to accommodate a package the branch doesn't even contain.

## 2. Playwright tested a build that no longer existed

The config serves a prebuilt `.next` and reuses an existing server outside CI. A test asserting a panel was collapsed by default kept failing *after* the fix was written and typechecked — the served bundle predated the edit.

The conclusions available at that moment were "the fix doesn't work" or "the `Collapsible` primitive is broken." The truth was "you're testing a stale artifact." Both wrong answers send you debugging the wrong layer.

## The guards

**`assert-install-matches-lockfile.mjs`** compares `package-lock.json` against the `node_modules/.package-lock.json` npm maintains — exact, not heuristic.

Optional and platform-gated entries are excluded. A lockfile lists every platform's native binaries (`@esbuild/android-arm`, `@rollup/rollup-linux-*`) and npm installs only this host's; treating those absences as drift made the check report **171 false positives on a clean tree**, which would train everyone to ignore it — the opposite of the point.

**`assert-build-is-current.mjs`** fails when any watched source is newer than the build manifest, naming the file and the gap:

```
.next is STALE: components changed 865s after the last build.
E2E would run against the previous build, so failures would describe code you have already changed.
```

## Wiring

Both are attached to the scripts that were misleading, not left as optional extras:

```
bundle:check  →  assert-install-matches-lockfile && check-bundle-budget
test:e2e      →  assert-install-matches-lockfile && assert-build-is-current && playwright test
```

Also exposed as `install:verify` and `build:verify` to run alone.

## Verification

Each guard tested in **both** directions — passing on a clean tree and failing on deliberately reproduced drift, including the exact case that caused this:

```
next: lockfile 16.2.12, installed 16.3.0
@next/env: lockfile 16.2.12, installed 16.3.0
```

33/33 E2E · `tsc` · `eslint` · bundle budget · `make preflight` all clean.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -10,11 +10,11 @@
     "dev": "next dev",
     "build": "next build --webpack",
     "start": "next start",
-    "bundle:check": "node scripts/check-bundle-budget.mjs",
+    "bundle:check": "node scripts/assert-install-matches-lockfile.mjs && node scripts/check-bundle-budget.mjs",
     "typecheck": "tsc --noEmit",
     "lint": "eslint",
     "lock:normalize": "npm install --package-lock-only --ignore-scripts",
-    "test:e2e": "playwright test",
+    "test:e2e": "node scripts/assert-install-matches-lockfile.mjs && node scripts/assert-build-is-current.mjs && playwright test",
     "verify:toolchain": "node scripts/verify-toolchain.mjs",
     "headers:sync": "node scripts/sync-vercel-headers.mjs",
     "headers:check": "node scripts/sync-vercel-headers.mjs --check",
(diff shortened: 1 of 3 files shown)


msaad00:
Combined into #4728 with the root-cause fix for the metrics-snapshot churn that was blocking both.

PR #4752, opened 2026-08-10
fix(release): the tag check must never wait for a credential prompt

msaad00:
**`Test (Alpine/musl)` hit its 45-minute ceiling on #4749 with no failing test
to point at.** pytest started at 19:43:48 and produced **no output at all**
before the job was killed at 20:28 — a hang, not a slow suite.

The cause is in the script I added in #4739. `_remote_tags()` runs

```
git ls-remote --tags origin
```

with **no timeout** and nothing to stop git from prompting. Against a private
remote in a container with no credentials, git blocks waiting for a username —
so a release-gate helper hung the entire job. The suite runs with `-x`, so
nothing after it ran either, which is why the log is empty rather than partial.

## Three guards, one principle: a check that cannot answer must fail, not wait

- `GIT_TERMINAL_PROMPT=0`, `GIT_ASKPASS=true`, `GCM_INTERACTIVE=never` — git
  fails immediately instead of asking a human who isn't there.
- **A 30-second timeout on every git call.** Generous enough that a slow network
  still answers; short enough that a wedged process fails the check, not the job.
- `TimeoutExpired` / `FileNotFoundError` / `OSError` are treated as "remote
  unreachable", which already prints the warning added in #4750 rather than
  silently reporting a shipped version as new.

**This matters beyond CI.** `preflight_release.sh` runs this on a release
engineer's machine, where a hung git looks like a frozen terminal during the one
procedure that should never be ambiguous.

## Verified in the failure mode itself

With a `git` shim on `PATH` that sleeps 600s on `ls-remote`:

- **before:** blocks indefinitely
- **after:** returns in **30.4s**, still reports the drift line, still refuses a
  shipped version

7 tests pass; `ruff`, `ruff format --check`, `mypy` clean.

Merging this unblocks the Alpine job on #4749.


msaad00:
Superseded — this change already shipped in #4749 (merged 21:54Z). The merge-from-main at 21:57Z brought main's copy back into this branch, which is why the diff reads as zero files.

Verified the fix is intact on main: `scripts/check_release_tag_is_new.py` is byte-identical on both refs (blob `d3e2c3dd4200c3de332bd9cc6423fc97a58a4673`) — `GIT_TERMINAL_PROMPT=0`/`GIT_ASKPASS`/`GCM_INTERACTIVE`, the 30s timeout, and the `TimeoutExpired`/`OSError` handling are all present. Nothing lost by closing.

PR #4762, opened 2026-08-11
fix(graph): a page must not list the same edge twice

msaad00:
## What

`GET /v1/graph` returned **more edges than the snapshot contains**. Reproduced on the demo estate: **486 duplicate edges**, every one a `contains` relationship. They aren't new relationships — they're the same relationships listed a second time, which inflates any count a client derives from the payload and makes an edge look like corroborating evidence for itself.

## Mechanism

`_containment_ancestors` begins its walk from `frontier_edges = known_edges` with `frontier = set(node_ids)`, keeping any edge where the **target is on the page and the source is not** — a containment edge reaching in from a parent outside the page.

But `known_edges` *is* the page's own edge list, fetched by `edges_for_node_ids`, which returns every edge touching a page node **regardless of which end is on the page**. So every such edge was already in `paged_edges`, and the route then serialised:

```python
"edges": [e.to_dict() for e in [*paged_edges, *ancestor_edges]],
```

a plain concatenation. This is only reachable when a page has a containment parent outside it, which is why it grows with estate size and stayed invisible on small fixtures.

## Fix

The join now keys on `(source, target, relationship)` — the same triple `UnifiedGraph.add_edge` already uses — so the response cannot hold a second, looser opinion about what makes an edge the same edge.

## Scoped deliberately

I checked the sibling sites rather than assuming they shared the bug:

- `graph.edges` appends **only past `add_edge`'s dedup check**, so it holds no duplicates.
- `_induced_subgraph` and the governance projection slice that already-clean list, so both were already correct.

Only the one site that concatenated two *independently fetched* lists was wrong, so only that one changed.

## How verified

- New test watched fail first with the real number: `486 duplicate edges in the page`, naming the offending triples.
- A second test asserts a page can never carry more edges than the snapshot total.
- A third asserts the dedup does **not** drop the ancestor context the join exists to carry — no node in the payload may be left without an edge. This is the regression the obvious fix (dropping ancestor edges entirely) would have caused.
- `make preflight` green; `make lint-ruff`, `make format-check`, `mypy` clean.
- `test_graph_api` + `test_demo_estate_bootstrap` running; will confirm before merge.

### DIFF
--- a/src/agent_bom/api/routes/graph.py
+++ b/src/agent_bom/api/routes/graph.py
@@ -830,6 +830,27 @@ async def _containment_ancestors(
     return list(nodes), list(ancestor_edges.values())
 
 
+def _joined_edges(*groups: list[UnifiedEdge]) -> list[UnifiedEdge]:
+    """Concatenate edge lists without listing the same edge twice.
+
+    The containment-ancestor walk starts from the page's own edge list, so an
+    edge reaching the page from a parent outside it is found there *and* stays
+    in ``paged_edges`` — concatenating the two returned it twice. The key is the
+    one ``UnifiedGraph.add_edge`` already uses, so the response cannot disagree
+    with the graph about what makes an edge the same edge.
+    """
+    seen: set[tuple[str, str, str]] = set()
+    joined: list[UnifiedEdge] = []
+    for group in groups:
+        for edge in group:
+            key = (edge.source, edge.target, _rel_value(edge))
+            if key in seen:
+                continue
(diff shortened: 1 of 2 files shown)


msaad00:
Local suites confirmed: `test_graph_api` + `test_demo_estate_bootstrap` + the new file — **110 passed** (10m16s). That was the one claim left open in the description.

agent-bom (approved):

andres-linero (approved):

PR #4763, opened 2026-08-11
fix(findings): "reachable" must name who reaches it

msaad00:
## What

`safe_finding_response_payload` sanitises a finding row on **two** paths — the API response, and persistence of compliance-hub findings via `_redact_finding`. It copied `graph_reachable` through after a type check and nothing else:

```python
graph_reachable = row.get("graph_reachable")
payload["graph_reachable"] = graph_reachable if isinstance(graph_reachable, bool) else None
```

So a caller could assert `graph_reachable: true` on a finding with no supporting path and collect `RISK_REACHABLE_BOOST` — **6.5 where the same finding without the claim scores 6.0**. That outranks findings the reachability engine actually proved, on the one signal that is supposed to mean *an agent can get here*.

## This is not a new policy

`dependency_reach.py` already defines the word:

```python
@property
def reachable(self) -> bool:
    return bool(self.reachable_from)
```

Reachable **means** at least one agent reaches it. And `blast_reach.py` shows the engine never emits the contradiction — when it stamps `reachable` it stamps the agent list and hop distance from the same report. The guard enforces the definition the codebase already holds rather than inventing one.

`_redact_finding` sharpens it: it *drops* `graph_reachable_from_agents` when empty while keeping `graph_reachable`, so the persisted row keeps the claim and discards the evidence for it. The new check runs **before** that drop.

## Fix

A `true` with no surviving agent id becomes `None` — the model's documented "engine did not run" state, which leaves scoring unchanged. An unsupported claim costs nothing rather than buying rank. Checked against the **sanitised** agent list, so unusable ids don't count.

## What I deliberately did NOT do

**`false` is still accepted.** It lowers risk and is equally unverifiable, so it's tempting to reject too — but a *genuinely* unreachable finding carries no corroborating fields either (the engine emits `distance=None`, `agents=[]`), making it indistinguishable from a fabricated `false`. Rejecting it would break round-trip of our own export. Separating them needs a provenance field asserting the engine ran, which this change does not add.

**A second, related inconsistency I found but did not silently resolve.** `api/finding_reachability.py:179` sets `graph_reachable = True` from attack-path evidence, but `_reaching_agents` only counts hops whose id starts with `agent:`. An internet→package path therefore yields `reachable=True` with an empty agent list — the same shape this PR rejects on ingest. That is a genuine conflict between two definitions of the word (reachable-by-agent vs reachable-by-any-path), and picking one is a product decision about what the risk boost should mean, not a bug fix. Flagging it rather than deciding it.

## How verified

- 12 new tests; 5 watched fail for product reasons first, including the score assertion failing `6.5 == 6.0` — so the boost being removed is demonstrated, not assumed. The test also asserts the boost is real, so the comparison isn't vacuous.
- `pytest -k "reachab or finding_scope or scope"` — 632 passed, 11 skipped.
- `make lint-ruff`, `make format-check`, `mypy` clean.

### DIFF
--- a/src/agent_bom/finding_scope.py
+++ b/src/agent_bom/finding_scope.py
@@ -628,8 +628,6 @@ def safe_finding_response_payload(row: Mapping[str, Any]) -> dict[str, Any]:
     if isinstance(row.get("is_kev"), bool):
         payload["is_kev"] = row["is_kev"]
 
-    graph_reachable = row.get("graph_reachable")
-    payload["graph_reachable"] = graph_reachable if isinstance(graph_reachable, bool) else None
     graph_distance = row.get("graph_min_hop_distance")
     payload["graph_min_hop_distance"] = (
         graph_distance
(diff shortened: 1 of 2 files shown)


msaad00:
Remaining local gates confirmed: compliance-hub suite **348 passed, 22 skipped** (the `_redact_finding` path this change runs ahead of), and `make preflight` green.

agent-bom (approved):

andres-linero (approved):

PR #4771, opened 2026-08-11
feat(findings): make KEV filterable, not just displayed

msaad00:
Closes the last open finding from the pre-publish audit.

## What

The executive view leads with a KEV count — *actively exploited, fix these first* is the strongest prioritisation signal the product has. `is_kev` is carried on every finding row and rendered per finding, but `/v1/findings` had **no parameter to filter on it**. The headline was a dead end: a reader could see that N findings are known-exploited and had no way to ask for those N.

## Where the predicate lives

In `row_matches_scope` — the single source of truth shared by the route's in-memory scan findings **and** the hub store's bulk-ingested keyset path. Putting it anywhere else would have created a second implementation of one judgement, which is the defect class this release has been burning down.

## Two deliberate choices

- **An absent `is_kev` is not a match.** A row nobody checked must never be presented as actively exploited — the same fail-closed rule the reachability (#4763) and completeness (#4760) work landed on.
- **`kev=false` is the real inverse**, not a no-op, and omitting the parameter changes nothing. A filter you can't turn off isn't a filter.

## Full-stack

Route parameter → worker-thread call → shared predicate → regenerated OpenAPI artifacts → `ui/lib/api.ts`. No half-wiring.

`make preflight` caught the OpenAPI drift on the first run; regenerated, and the diff is only the `kev` parameter.

## How verified

- 10 tests, watched fail first (3 red for product reasons — the absent-verdict and inverse cases).
- **End-to-end on the demo estate**: 63 KEV findings, every returned row `is_kev: true`, inverse excludes all 63.
- `pytest -k "kev or finding_scope or findings"` — exit 0.
- `make preflight`, `make lint-ruff`, `make format-check`, `mypy` clean; `npm run typecheck` clean.

One note on process: an earlier run of that suite reported 83 failures. It had been launched **mid-edit**, while `kev` was still undefined at a call site (ruff flagged it as F821 moments later). I re-ran it against the finished code rather than reason the number away — it passes.

### DIFF
--- a/docs/openapi/v1.json
+++ b/docs/openapi/v1.json
@@ -17165,6 +17165,24 @@
             }
           },
           {
+            "description": "Only known-exploited (KEV) findings, or only non-KEV when false",
+            "in": "query",
+            "name": "kev",
+            "required": false,
+            "schema": {
+              "anyOf": [
+                {
+                  "type": "boolean"
+                },
+                {
+                  "type": "null"
+                }
+              ],
+              "description": "Only known-exploited (KEV) findings, or only non-KEV when false",
+              "title": "Kev"
+            }
+          },
+          {
             "in": "query",
             "name": "include_facets",
             "required": false,
--- a/docs/openapi/v1.yaml
+++ b/docs/openapi/v1.yaml
@@ -11544,6 +11544,16 @@ paths:
             type: string
           - type: 'null'
           title: Finding Class
+      - description: Only known-exploited (KEV) findings, or only non-KEV when false
+        in: query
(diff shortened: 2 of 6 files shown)


msaad00:
Test counts for the re-run referenced in the description: **667 passed, 7 skipped, 0 failed** (`pytest -k "kev or finding_scope or findings"`). Same selection that reported 83 failures mid-edit — that state had `kev` undefined at a call site.

agent-bom (approved):

andres-linero (approved):

PR #4772, opened 2026-08-11
chore(release): prepare 0.100.0 — changelog, version, generated artifacts

msaad00:
**Draft — one step remains (screenshots). Do not merge yet.**

## Version: 0.100.0, not 0.99.1

A **minor** bump. Since v0.99.0 this adds a public REST endpoint (`GET /v1/findings?kev=`), a scan remediation endpoint, and an MCP OAuth authorization-server surface. Patch would understate it. 46 commits since the tag.

## What's in

- **CHANGELOG entry for 0.100.0** — written for a reader, not as a commit list. Grouped into Security / Fixed — wrong or missing output / Added / Changed, leading with the RHEL-UBI false clean because it's the one that changes what an existing user sees.
- **Version bump across the tree** — 71 files.
- **Regenerated, not hand-edited**: OpenAPI artifacts and the agent capability manifest both embed the version and both drifted. `bump-version.py` alone does not cover them — `make preflight` caught each in turn, which is the gate working.

## What remains — and why I did NOT do it

`docs/images/product-screenshots.json` still declares `0.99.0`, so **preflight currently fails**:

```
ERROR: docs/images/product-screenshots.json has stale release_version: '0.99.0' != 0.100.0
```

The fix is **not** to edit that version string. The screenshots must be re-captured against the new build via `ui/scripts/capture-product-proof.mjs`, which needs a built UI and a live API with the demo estate. Editing the string satisfies the gate while shipping images from the previous release — exactly the failure the gate exists to catch. I'd rather hand you a red draft than a green lie.

## Correction to something I reported earlier

I stated several times that **main has never had a fully green CI run**. That is **wrong** — I carried it from the audit notes without checking. Verified: run `31535310585` on main is **10 success / 9 skipped / 0 failed** (the skips are path-filtered jobs behaving correctly). Main's other runs show `cancelled`, which is concurrency cancellation from rapid successive merges, not failure. Main is healthy.

## Release state

All 18 audit fixes are merged (#4753–#4771). Two items are escalated to the owner rather than silently decided:

1. **Two definitions of "reachable"** (by-agent vs by-any-path) — decides what `RISK_REACHABLE_BOOST` means.
2. **Structured remediation is tier-A blocked** by a deliberate data-leakage policy; surfacing it needs a provenance distinction, not a patch.

Also still open: Smithery is stale at 36/77 tools and needs `AGENT_BOM_OAUTH_AS_PRIVATE_KEY_PEM` on the hosted deployment.

Tagging and publishing are yours — I have not tagged.


msaad00:
## Screenshot recapture — attempted, not finished. Here's exactly where it stands.

I ran the capture rather than edit the version string, and got most of the way. **The manifest still reads `0.99.0`, so this PR is still correctly red.**

### What worked
The harness boots the committed `.next/standalone` build and captured a series of surfaces cleanly (`gateway-policies-live.png`, `identity-audit-live.png`, `dependency-map-live.png`, and others).

### Where it stops
```
Error: Expected content DEMO-VULN-21441 is missing on /remediation?capture=1
```

**This is a data-seeding mismatch, not a page regression.** I checked the running API directly: it had findings, but zero `DEMO-VULN-*` ids — the sample was real CVE data (`CVE-2020-14343: pyyaml@5.3`). So the page is fine; it was pointed at the wrong dataset.

### My mistake, so the next run doesn't repeat it
I started the API with `agent-bom serve --demo-estate`. That is **not** the documented protocol. `docs/CAPTURE.md` specifies a **fresh SQLite database**, then generating the demo report via `agent-bom scan --demo --offline -f json` and pushing **that exact payload** — which is where the `DEMO-VULN-*` fixtures come from. `--demo-estate` seeds a different set entirely.

Next run should follow `docs/CAPTURE.md` from step one rather than my shortcut.

### One thing worth noting beyond this PR
My first capture attempt reported **exit code 0 while failing**, because the command was piped to `tail`. That is the second time today a green exit masked a real failure (the first was `npm install --silent` hiding an ERESOLVE conflict). If any CI step pipes a build or capture through another command without `set -o pipefail`, it can report success on failure — worth a look independently of the release.

msaad00:
## Fixed — screenshots recaptured, preflight green, out of draft

`docs/images/product-screenshots.json` now declares **0.100.0** with images actually captured from this build. `make preflight` passes.

### The real defect, which was mine

The harness had **no mock for `/v1/scan/{job_id}/remediation`** — the endpoint added earlier in this cycle. The remediation page fell through to whatever was on the API port and rendered without its content. This is the same gap #4737 closed for the e2e suite; the capture harness was missed. The plan is now served from `scanJob().result.remediation_plan` rather than a second fixture, so the mocked endpoint and the mocked job cannot disagree.

`DEMO-VULN-21441` is a fixture **inside the harness**, so no live API was ever required — my first three attempts were chasing the wrong thing.

### Four failures, each real and worth recording

| # | What I did | What actually happened |
|---|---|---|
| 1 | no API running | `/remediation` 500ed on ECONNREFUSED |
| 2 | `agent-bom serve --demo-estate` | seeded a different dataset than the harness expects |
| 3 | `--allow-insecure-no-auth` | results push **403'd** — anonymous gets `viewer`, `/v1/results/push` needs `analyst` |
| 4 | uncommitted tree | harness refused to stamp a release manifest from a dirty tree |

Two of those are guards working exactly as designed (#3 role separation, #4 clean-tree enforcement), and I'd rather they fired than not.

### Also fixed
The `[Unreleased]` compare link had to start at `v0.100.0`, with a `[0.100.0]` link added — caught by the release-consistency gate.

### Still worth a separate look
My first capture attempt reported **exit code 0 while failing**, because it was piped to `tail`. Second time today a green exit masked a real failure (the first was `npm install --silent` hiding an ERESOLVE conflict). Any CI step piping a build through another command without `set -o pipefail` can report success on failure.

andres-linero (approved):

andres-linero (approved):

agent-bom (approved):

PR #4780, opened 2026-08-12
chore(release): the release run owns the Docker MCP submission pin

msaad00:
Closes the first of the two open pre-tag decisions: **the release run should own `server.yaml` the way it now owns `sdks/python`.**

## The problem the pin has that the version bump does not

`integrations/docker-mcp-registry/server.yaml` tells Docker which commit to build the published server from:

```yaml
source:
  commit: cc4640f10e4428f0c76ff989e0dd7bdf5c159239
```

That value is the commit a release tag points at — so unlike every other version-bearing artifact, it **cannot be written before the tag exists**. `bump-version.py` runs before tagging and structurally cannot own it. So it was carried as prose instead:

> `SUBMISSION.md` step 1: Update `source.commit` in `server.yaml` to the new release tag SHA

That is exactly the shape #4774 gated everywhere else — a value a human is asked to remember — and it drifted the same way: the pin has not moved in several releases.

## What now owns it

A `docker-mcp-pin` job runs after `github-release`, writes `${{ github.sha }}` into the pin and opens `chore/docker-mcp-pin-vX.Y.Z`.

- **Branch + PR, not a push to `main`** — matching `mcp-registry-sync.yml` and `uv-lock-upgrade.yml`, the repo's existing convention for CI-authored changes, and the same SSH-signed automation commit with the same graceful skip when `AUTOMATION_SSH_SIGNING_KEY` is unset.
- **`continue-on-error: true`, gated on `refs/tags/v`** — by the time it runs, PyPI, Docker Hub, Helm and the GitHub Release are already published. A pin PR that fails to open must not turn a shipped release red. A test asserts this, so it cannot be removed quietly.

## What CI gates — and what it deliberately does not

`sync_docker_mcp_pin.py --check` asserts the pin is a full 40-character SHA that is a real commit in this repository (a prefix or a branch name fails; a shallow clone reports unverifiable rather than failing).

**Freshness is deliberately not gated.** The release run moves the pin by PR *after* the tag exists, so between the tag and that merge the pin is legitimately one release behind. A freshness gate would fail every release — the precise reason to say so here rather than discover it during the next one.

## Verified

Each path was watched failing before it was made to pass:

| input | result |
|---|---|
| `--set deadbeef` | rejected — not a full SHA |
| `--set <real sha>` | `cc4640f1…` → `8675f48e…`, one line changed |
| `--set` same SHA again | no-op |
| pin truncated to `cc4640f` | `--check` fails |
| pin set to 40 zeroes | `--check` fails — not a commit here |

Four mutations, each confirmed to fail the guard that claims to catch it: deleting the `docker-mcp-pin` job, removing `continue-on-error`, reverting SUBMISSION.md to the manual step, and making `write_pin` rewrite every `commit:` key rather than only the one under `source:`.

Also green: `lint_release_workflow.py` (the release call-graph guard), `check_ci_pipefail.py`, `check_release_consistency.py`, `check_published_counts.py`, ruff, mypy.

## One thing found while writing this

My first implementation matched the pin with a `(?ms)` regex. Under `DOTALL`, `.` inside the lazy repeat swallowed newlines and the match backtracked catastrophically — it hung for over two minutes on a 47-line file rather than failing. Replaced with a line scanner, which is both bounded and easier to read. Flagging it because a ReDoS-shaped regex in a release-path script is worth not repeating.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -210,6 +210,12 @@ jobs:
       - name: Published counts match the shipped build
         run: python scripts/check_published_counts.py
 
+      # Format and reachability only, deliberately not freshness: the release
+      # run moves the pin by PR *after* the tag exists, so between the tag and
+      # that merge the pin is legitimately one release behind.
+      - name: Docker MCP submission pin is a real commit
+        run: python scripts/sync_docker_mcp_pin.py --check
+
       - name: Docker base-image policy
         if: needs.changes.result != 'success' || needs.changes.outputs.docs_only != 'true'
         # Source: scripts/check_docker_base_policy.py
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1333,6 +1333,98 @@ jobs:
             echo "Updated floating tag $MAJOR → ${{ github.ref_name }}"
           fi
 
+  docker-mcp-pin:
+    name: Sync Docker MCP submission pin
+    # The pin must name the commit the tag points at, which does not exist
+    # until the tag does — so it cannot be written by bump-version.py with the
+    # other version-bearing artifacts. The release run owns it here instead of
+    # SUBMISSION.md asking a human to remember. Never blocks the release: the
+    # artifacts are already published by the time this runs.
+    needs: github-release
+    if: startsWith(github.ref, 'refs/tags/v')
+    runs-on: ubuntu-latest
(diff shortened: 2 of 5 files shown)


msaad00:
Superseded by #4783, which carries this commit unchanged plus the other half of the pre-tag work. Closing so there is one PR to approve instead of two.

PR #4781, opened 2026-08-12
docs(changelog): record the wave the 0.100.0 entry was written before

msaad00:
**Merge this last.** It describes #4773–#4778 and #4780, so it should land after them.

## Why it was wrong

The `0.100.0` entry was written and merged in #4772 — *before* the audit wave. Everything merged after it was therefore absent from the release notes, and the owner's decision is that all of it ships in 0.100.0. Tagging on the #4772 text would have understated the release in exactly the direction that matters: it omitted the security findings.

Missing before this PR:

| Missing | From |
|---|---|
| Two release gates that could never fail | #4774 |
| Postgres CI-vs-Docker schema divergence | #4774 |
| Compose bind mount resolving to a missing directory | #4774 |
| Severity filter that deleted the topology | #4776 |
| SQLite budget applied after materialization — an OOM path | #4776 |
| Demo-deploy gate asserting a 20-asset estate against 3,860 | #4773 |
| Advertised registry counts stale on six surfaces, two at runtime | #4777 |
| `sdks/python` frozen at 0.92.0 | #4777 |
| A documented first-run flag that never existed | #4777 |
| Canvas growing to 21,604px; graph text alternative; offline import | #4775 |
| Exit-code contract, `SUPPORT.md` | #4778 |
| Release run owning the Docker MCP pin | #4780 |

## Two structural changes

**The `[Unreleased]` block is folded into `0.100.0`.** #4776 filed its five graph entries under `[Unreleased]`, which was right when written and wrong once the owner decided the whole wave ships in this release. `[Unreleased]` is now empty.

**The intro says what the release turned out to be.** It already said "an output-honesty release … nothing here fixes a crash or a build". That stayed true, but it now also names the gates, because a release whose headline finding is *our own security gate could not fail* should not bury it under Fixed.

## On the dead-gate entry specifically

Written to state both halves, because either alone misleads:

> the gate could not have reported a vulnerable dependency — and, run for real against all three lockfiles, it reports none, so nothing was shipped vulnerable.

Omit the first half and it hides that prior releases passed the gate vacuously. Omit the second and it implies we shipped something vulnerable. We did not.

## Verified

- The release workflow's `version-guard` regex `^## \[0.100.0\](?:\s|$)` still matches — checked directly, since this PR moves that header.
- `[Unreleased]` parses as empty.
- `tests/test_release_tag_is_new.py` — 7 passed.
- Compare links at the bottom of the file are untouched.

Every entry is drawn from the merged PR that shipped it, not paraphrased from the audit that requested it — measured numbers (270MB→6.4MB, 21,604px→680px, 0.064→0.380, 1,013 vs 1,081, 940 vs 60) are the ones those PRs recorded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,59 @@ Versions follow [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
-### Fixed
+## [0.100.0] - 2026-08-11
+
+An output-honesty release. Every local gate was already green before this work
+started; nothing here fixes a crash or a build. What it fixes is the gap between
+what agent-bom found and what it told you — findings that were computed and never
+surfaced, bounded reads that claimed to be complete, a scanner that reported
+clean where it had not looked, and two gates on the release path that could not
+have failed if there had been something to catch.
+
+If you run agent-bom against RHEL/UBI images, or read the executive view, or use
+the graph, this release changes what you see.
+
+### Security
+
+- **Two gates on the release path could never fail.** GitHub runs an unqualified
+  `run:` step as `bash -e {0}` — without `-o pipefail` — so `osv-scanner … | tee`
+  exited with `tee`'s status and the `|| { … }` fixable-CVE branch below it was
+  unreachable. That dead branch was the blocking logic in `ci.yml` *and* in
+  `release.yml`, where the step is named "block release on fixable CVEs". Every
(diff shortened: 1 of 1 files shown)


msaad00:
Superseded by #4783, which carries this commit unchanged plus the other half of the pre-tag work. Closing so there is one PR to approve instead of two.

PR #4782, opened 2026-08-12
feat: add optional You.com search intel tool

mouse-value-add:
This adds an optional You.com-backed search tool to the existing threat-intel MCP surface.

Why this fits here:
- `agent-bom` already has threat-intel, advisory, and daily-brief tools, so a live search helper is a natural extension for current context and news.
- I kept it opt-in: the tool only works when `YDC_API_KEY` is set.

What changed:
- Added `youcom_search` to the MCP server and server-card metadata.
- Implemented a compact You.com Search API helper that returns web/news results as JSON.
- Documented the new tool in the README and updated the MCP tool count.
- Added tests for the missing-key path and a mocked success path.

Setup:
- `YDC_API_KEY` enables the tool.
- `YOUCOM_BASE_URL` can override the default You.com host.

Usage example:
```text
agent-bom mcp server
# then call youcom_search from an MCP client
```

Fallback / error handling:
- If `YDC_API_KEY` is missing, the tool returns a clear validation error.
- Upstream API failures return a structured upstream-unavailable error.
- Existing default behavior is unchanged for users who do not set the new env var.

Validation:
- `uv sync --extra dev-all`
- `uv run python -m pytest tests/test_intel_lookup.py tests/test_mcp_server.py -q`
- `uv run ruff check src/agent_bom/mcp_tools/intel.py src/agent_bom/mcp_server.py src/agent_bom/mcp_server_metadata.py tests/test_intel_lookup.py tests/test_mcp_server.py`

Tracking issue:
- https://github.com/youdotcom-oss/integration-tracking/issues/185

### DIFF
--- a/PROJECT_STRUCTURE.md
+++ b/PROJECT_STRUCTURE.md
@@ -87,7 +87,7 @@ The self-hosted operator surface. Same evidence, multi-tenant, audited.
 | Subsystem | Path | Responsibility |
 |---|---|---|
 | API | `api/` | FastAPI app and route modules; middleware handles HTTP auth/tenant/limits/audit. SQLite and Postgres are the primary stores, Neptune is optional graph persistence, ClickHouse is optional analytics, and Snowflake implements selected store/warehouse paths. |
-| MCP server | `mcp_server*.py`, `mcp_tools/` | FastMCP server advertising 77 tools, 6 resources, and 8 workflow prompts (mostly read-only; Shield write actions fail closed). |
+| MCP server | `mcp_server*.py`, `mcp_tools/` | FastMCP server advertising 78 tools, 6 resources, and 8 workflow prompts (mostly read-only; Shield write actions fail closed). |
 | Runtime enforcement | `proxy*.py`, `gateway*.py`, `firewall*.py`, `shield.py`, `runtime/`, `enforcement.py` | MCP traffic proxy, secure-by-default gateway, inline firewall, Shield enforcement. **Spread by design** — see [Runtime enforcement spread](#runtime-enforcement-spread). |
 | Auth / tenancy | `rbac.py`, `permissions.py`, `entitlements.py`, `mcp_tenant.py`, `api/auth.py`, `api/oidc.py` | RBAC roles, tenant scoping, API keys, OIDC/SAML/SCIM. |
 | Fleet | `fleet/`, `fleet_scan.py` | Endpoint/collector inventory pushed into one control plane. |
--- a/PYPI_README.md
+++ b/PYPI_README.md
@@ -115,7 +115,7 @@ helm upgrade --install agent-bom deploy/helm/agent-bom \
 - **Blast radius + attack-path fusion** — multi-hop exposure paths over one `UnifiedGraph`, from package → finding → MCP server → credentials → connected agents
 - **CWE-aware impact** — RCE shows credential exposure, DoS does not; symbol-level CVE reachability joins CWE/CVE/CPE advisory context when OSV/GHSA carry affected symbols (Python, npm, Go)
 - **Portable outputs** — SARIF, CycloneDX, SPDX, OCSF, HTML, graph, JSON, ZIP evidence bundles, and more
-- **MCP server mode** — 77 MCP tools, 6 resources, and 8 workflow prompts exposed to MCP clients like Claude, Cursor, Windsurf, and Cortex CoCo / Cortex Code
+- **MCP server mode** — 78 MCP tools, 6 resources, and 8 workflow prompts exposed to MCP clients like Claude, Cursor, Windsurf, and Cortex CoCo / Cortex Code
 - **Skill bundle identity** — stable bundle hashes for skill and instruction file review
 - **Dependency confusion detection** — flags internal naming patterns
 - **VEX generation** — auto-triage with CWE/CVE/CPE-aware reachability (package + function-level when advisory symbols exist)
--- a/README.md
+++ b/README.md
@@ -184,11 +184,15 @@ real identity, TLS, PostgreSQL, encryption, and audit keys before exposing it.
 | GitHub CI | `uses: msaad00/agent-bom@v0.100.0` | SARIF, PR summary, and a policy exit code |
 | Cloud evidence | `agent-bom connect aws` | Stored connection reference; run scans from the control plane |
 | Runtime gateway | `agent-bom gateway serve --from-control-plane http://127.0.0.1:8422 --bind 127.0.0.1:8090` | Allow, warn, and block audit events |
-| Agent interface | `agent-bom mcp server` | 77 MCP tools, 6 resources, and 8 workflow prompts |
+| Agent interface | `agent-bom mcp server` | 78 MCP tools, 6 resources, and 8 workflow prompts |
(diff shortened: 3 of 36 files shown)


msaad00:
Thanks for this — it's a genuinely well-formed first contribution. It uses the right internals (`mcp_error_json`, `_execute_tool_async`, `_READ_ONLY`, `sanitize_error`), it's opt-in, and it ships tests for both the missing-key and success paths. Two things I expected to find wrong and didn't: `create_client()` already calls `check_offline()` and disables redirects, so offline mode and redirect-based SSRF are handled for free.

We're taking it. I'm going to finish it on your branch rather than send you round-trips, since most of what's left is repo-specific knowledge you had no way to have. Writing it all down so the review is auditable.

## No CI ran on this

First-time-contributor workflows need maintainer approval, so nothing in the checks list executed. I ran the gates locally against `4b620ce94`.

## 1. The tool count is advertised on 15 surfaces, not 1

This is the main thing, and it's not your fault — it's a defect class we gated only in the last two days (#4774, #4777). Adding a tool moves a number that is asserted from the built server card:

```
$ python scripts/check_published_counts.py
ERROR: 11 published count(s) disagree with what the build ships:
  - PYPI_README.md:118                        advertises 77, build ships 78
  - docs/CLAUDE_INTEGRATION.md:62             advertises 77, build ships 78
  - docs/MCP_SERVER.md:3, :168                advertises 77, build ships 78
  - docs/PRODUCT_BRIEF.md:134                 advertises 77, build ships 78
  - docs/images/persona-value-{dark,light}.svg:21   advertises 77, build ships 78
  - integrations/docker-mcp-registry/SUBMISSION.md:8
  - integrations/smithery.yaml:6
  - site-docs/getting-started/mcp-server.md:3
  - ui/app/connections/page.tsx:2812
```

plus `scripts/check_release_consistency.py` exit 1, and five failures in `tests/test_stats_alignment.py` covering `integrations/docker-mcp-registry/tools.json`, `src/agent_bom/cli/_server.py` and `deploy/docker/Dockerfile.sse`.

Your validation command (`pytest tests/test_intel_lookup.py tests/test_mcp_server.py`) was real, just narrower than the blast radius. For future reference the two that catch this are `scripts/check_published_counts.py` and `tests/test_stats_alignment.py`.

`tools.json` needs the actual tool entry, not just a count bump.

## 2. `YOUCOM_BASE_URL` sends the API key to an unvalidated host

This is the one substantive security change I'm asking for. Today:

```python
base_url = (os.environ.get("YOUCOM_BASE_URL") or _YDC_BASE_URL).rstrip("/")
...
headers={"X-API-Key": api_key, ...}
```

There's no scheme or host check, so `YOUCOM_BASE_URL=http://attacker.tld` sends the user's key to an arbitrary origin in cleartext. It's an env var rather than remote input, so it's not remotely exploitable — but this repo has fixed this exact class twice (#4626, the Docker Hub bearer token following a redirect off-origin; and Microsoft Graph pagination in `cloud/azure_graph.py:47-65`), and a credentialed request to a configurable host is the shape we now pin by default.

The pattern to follow is `ai_enrich.py:328`, which allows the override only when it is loopback. So: require `https` and pin to the You.com host, unless the override resolves to loopback (which keeps your test/dev use case working).

## 3. The Docker MCP distribution would ship it broken

`integrations/docker-mcp-registry/server.yaml` has neither `ydc-index.io` in `allowHosts` nor `YDC_API_KEY` in `config.secrets`. The tool is advertised into that distribution by `tools.json` but egress is blocked there, so it would fail for every Docker Desktop user.

## Two smaller notes

- `response.text[:500]` is echoed into the error `details` on a non-200. Everything else in this file goes through `sanitize_error`; an upstream body shouldn't reach an MCP client unfiltered.
- `count` is clamped 1–100 and `crawl_timeout` 1–60, but `freshness`, `country`, `language`, `safesearch` and `livecrawl` are forwarded as-is. Not a vulnerability — the upstream will reject them — but a bad value produces an opaque upstream error rather than a validation one.

## What happens next

I'll push the count sweep, the origin pin, the `allowHosts`/`secrets` entries and the sanitised error onto this branch, then run the full gate set. Your commit stays as the first commit — the feature and its tests are yours.

One thing worth stating plainly since it's the reviewer's job and not a criticism of the patch: this adds a third-party egress path from a surface that is otherwise local-first, so the queries an agent constructs (package names, CVE identifiers, hostnames) leave the customer's environment. That's why it stays opt-in behind `YDC_API_KEY`, why the host is getting pinned, and why the README line says explicitly what enabling it does.

msaad00:
Pushed the completion to this branch as `c8c877f75` — your commit is untouched underneath it. Everything below was run against the merged result, not asserted.

**Security — the one that mattered.** `YOUCOM_BASE_URL` is now resolved through `_resolve_youcom_base_url()`: honoured on a You.com origin over TLS, or on loopback so your local testing still works, and rejected otherwise before any request is made. Four rejection cases are covered by a parametrised test whose fake client *raises if it is ever called*, so the assertion is "the key never left", not merely "an error was returned":

```
https://attacker.example        → rejected
http://ydc-index.io             → rejected (cleartext)
https://ydc-index.io.attacker.example → rejected (suffix, not the origin)
ftp://ydc-index.io              → rejected
http://127.0.0.1:8931           → allowed
(unset)                         → https://ydc-index.io/v1/search
```

Mutation-tested: restoring the original `base_url` line fails all four.

**A real bug your types caught.** `results.get("web")` returns `None` when You.com omits a section rather than returning it empty, and that `None` was iterated — so a no-results query would have surfaced as `INTERNAL_UNEXPECTED` instead of an empty result set. Now `or []`. mypy found this, not review.

**Error path.** The non-200 body goes through `sanitize_error` before reaching an MCP client, since an upstream body can echo the request and the request carries the key.

**The count sweep.** 77 → 78 across all 15 surfaces, the `tools.json` entry added in place, and the deliberate tripwire in `check_release_consistency.py` bumped. Two of those surfaces are *generated* — `docs/images/architecture-{dark,light}.svg` derive the count from the server card — so they were regenerated rather than edited. Worth knowing if you contribute here again: hand-editing those two would have passed review and failed CI.

**Docker MCP distribution.** `ydc-index.io:443` added to `allowHosts` and `YDC_API_KEY` to `config.secrets`, so the tool can actually reach the network in the distribution `tools.json` advertises it into.

**Result:** 269 passed, 2 skipped across the MCP, stats-alignment, catalog-drift, deployment, surface-freshness, SVG and docs suites; `check_published_counts.py`, `check_release_consistency.py` and `check_ci_pipefail.py` all exit 0; ruff and mypy clean.

This ships in 0.100.0 and is credited to you in the CHANGELOG. Thanks — the tool is a genuine addition to the threat-intel surface, and the parts that needed changing were repo-specific conventions you had no way to know about.

msaad00:
Re-triggering CI: the `Update branch` merge commit does not fire workflow runs in this repo (a known quirk here — an API/merge-button push produces no `pull_request` event), so this branch currently has **zero** checks despite being up to date with `main`. Closing and reopening to generate a run, then approving it. No change to the code.

msaad00 (dismissed):

andres-linero (dismissed):

andres-linero (approved):

agent-bom (approved):

msaad00 (approved):

PR #4899, opened 2026-08-22
fix(demo): name the scan that owns the graph when seeding declines

msaad00:
## Summary
- record which snapshot the graph read path will actually serve when showcase seeding declines, on the bootstrap summary
- keep the exact graph-owner identifier out of logs; the completion log records only whether an owner identifier is present
- keep a preserved operator scan as the default while making fixed-timestamp showcase snapshots addressable by scan ID
- compact the graph evidence band without hiding freshness, completeness, or provenance signals
- keep estate-scale edge labels on the focused path while preserving all labels in capture mode
- record completed showcase attack-path analysis only after the final persisted path set is known
- derive findings severity chips from the facet contract so `info` and future buckets cannot be stranded

## Why
Showcase seeding refuses to shadow a real scan — correct, and covered by `test_explicit_demo_force_never_replaces_a_non_showcase_snapshot`, which states that an environment flag is not authorization to erase tenant graph history. But findings, identities and gateway events still seed, so a developer who ran `agent-bom scan .` before `--demo-estate` gets 3,469 estate findings beside a 15-node graph of their local repo, with nothing explaining the mismatch.

Measured on a seeded instance: `/v1/graph` returned 15 nodes / 26 edges (no tool, credential, identity, role or policy nodes; 2 `depends_on` edges) while the skipped showcase graph builds 99 nodes / 108 edges including 26 TOOL, 8 CREDENTIAL, 8 EXPOSES_CRED and 8 CAN_ACCESS.

The exact owner remains available in the bootstrap summary for callers that need it. It is not copied into application logs, satisfying the repository's sensitive-log boundary and the CodeQL data-flow gate.

The graph UI keeps freshness visible in a 26.5px evidence row and moves the canvas to roughly 546px at 1280×720 with a seeded 6,802-node estate. Dense graphs suppress unfocused edge captions; selected paths and capture mode retain their labels.

## Scope
The preservation boundary remains intact: no real snapshot is deleted or displaced. Showcase snapshots use their fixed older timestamps, so the newest-wins read path continues to serve the operator scan by default.

## Verification
- `UV_CACHE_DIR=/tmp/agent-bom-uv-cache uv run pytest -q tests/test_demo_estate_bootstrap.py tests/test_showcase_graph_seeding.py` — 56 passed
- `UV_CACHE_DIR=/tmp/agent-bom-uv-cache uv run ruff check src/agent_bom/demo_estate/bootstrap.py tests/test_demo_estate_bootstrap.py tests/test_showcase_graph_seeding.py`
- `UV_CACHE_DIR=/tmp/agent-bom-uv-cache uv run ruff format --check src/agent_bom/demo_estate/bootstrap.py tests/test_demo_estate_bootstrap.py tests/test_showcase_graph_seeding.py`
- `npm run verify:toolchain` — Node 24.19.0 / npm 10.9.7 contract verified
- `npm run typecheck`
- `npm test -- --run tests/findings-severity-buckets.test.ts tests/graph-utils.test.ts` — 14 passed
- `npm run test:e2e -- --grep "large estates lead with non-overlapping clusters"` — 2 passed (light and dark)
- `npm run build`
- local seeded browser QA at 1280×720 in light and dark themes — 6,802-node estate, compact evidence row, readable canvas

### DIFF
--- a/src/agent_bom/demo_estate/bootstrap.py
+++ b/src/agent_bom/demo_estate/bootstrap.py
@@ -428,10 +428,10 @@ def maybe_bootstrap_demo_estate(*, tenant_id: str = SHOWCASE_TENANT) -> dict[str
             }
         )
         _logger.info(
-            "demo estate bootstrap complete tenant=%s graph_seeded=%s graph_owner_scan_id=%s job_id=%s findings=%s",
+            "demo estate bootstrap complete tenant=%s graph_seeded=%s graph_owner_scan_id_present=%s job_id=%s findings=%s",
             tenant_id,
             graph_seeded,
-            summary.get("graph_owner_scan_id") or "-",
+            bool(summary.get("graph_owner_scan_id")),
             job_id,
             summary.get("findings"),
         )
--- a/src/agent_bom/demo_estate/bootstrap.py
+++ b/src/agent_bom/demo_estate/bootstrap.py
@@ -266,6 +266,17 @@ def _store_demo_scan_job(report: dict[str, Any], *, tenant_id: str) -> str:
     return job.job_id
 
 
+def _graph_owner_scan_id(graph_store: Any, tenant_id: str) -> str:
+    """Return the scan id whose snapshot the graph read path will serve."""
+    try:
+        latest = getattr(graph_store, "latest_snapshot_id", None)
+        if callable(latest):
+            return str(latest(tenant_id=tenant_id) or "")
+    except Exception:  # noqa: BLE001 - never block the demo boot on a read
+        _logger.debug("could not resolve graph owner scan id", exc_info=True)
+    return ""
+
+
 def maybe_bootstrap_demo_estate(*, tenant_id: str = SHOWCASE_TENANT) -> dict[str, Any]:
     """Seed showcase graph + curated findings when demo estate mode is enabled."""
     if not demo_estate_enabled():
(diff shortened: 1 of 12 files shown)


agent-bom (dismissed):

andres-linero (dismissed):

agent-bom (dismissed):

andres-linero (dismissed):

agent-bom (dismissed):

msaad00:
Current-head audit on `f5ccde538` found two remaining graph contract regressions:

1. **The dense-edge label budget is undone by the accessibility pass.** `baseDisplayEdges` calls `readableGraphEdges(..., { maxLabeledEdges: GRAPH_EDGE_LABEL_BUDGET })` and correctly removes labels above the budget. `accessibleBaseDisplayEdges` then calls `readableGraphEdges(baseDisplayEdges, undefined, { preserveVisualStyle: true })` with the default infinite budget. Because `exposes_cred` and the other high-signal relationships synthesize a label when none exists, the second pass restores every label. A 200-edge regression passed the first assertion (all labels removed), then failed after the accessibility pass (labels present again). The regression should exercise the two-pass graph-page pipeline; the implementation can preserve an existing label state when `preserveVisualStyle` is true, or otherwise carry the density/focus contract through the second pass.

2. **Forced showcase seeding can displace older real evidence.** With a preserved non-showcase snapshot at `2026-07-01T00:00:00+00:00`, `seed_showcase_graph_if_empty(..., force=True)` adds the fixed `2026-08-15` showcase snapshot and `latest_snapshot_id()` becomes `showcase`. The real snapshot is not deleted, but it is demoted from the default read path, contradicting `test_explicit_demo_force_never_replaces_a_non_showcase_snapshot` and the docstring's “never shadowed” contract. The current regression covers only a real scan newer than the fixed showcase stamp. Add the older/imported-snapshot case and make forced seeding addressable without changing the default owner.

msaad00:
Disposition: both findings are recorded for a separate, cohesive graph/demo truth PR after #4899 merges. They are not requests to expand or race the currently active #4899 branch.

andres-linero (approved):

agent-bom (approved):

PR #5020, opened 2026-08-30
fix(runtime): sanitize red-team provider failures

### LINKED ISSUE #5013: fix(runtime): sanitize red-team provider failures before logging
## Problem

src/agent_bom/red_team_llm.py logs a raw provider exception during best-effort variant generation. Provider exceptions may contain endpoints, request metadata, credentials, or response content.

## Acceptance criteria

- route log text through the canonical sanitization helper or emit only a stable error code/type
- never include raw provider exception text in logs or returned results
- add a regression with secret-like exception content and assert it is absent
- extend the exception-sanitization guard if the parameterized logger form is currently missed

## Verification

Run the focused red-team tests and scripts/check_exception_sanitization.py.

msaad00:
Summary
- replace raw red-team provider exception logging with a stable exception category
- extend the exception-sanitization guard to the red-team provider path
- add a secret-like provider failure regression test

Verification
- uv run pytest -q tests/test_red_team_llm.py tests/test_exception_sanitization_guard.py
- uv run ruff check on changed Python files
- uv run ruff format --check on changed Python files
- python scripts/check_exception_sanitization.py
- git diff --check

Closes #5013


msaad00:
@andres-linero All 41 checks are green at signed SHA 01533a28a3e4a86a0e8a0174945c3c64d182c594. Independent approval is the remaining protected-branch gate; squash auto-merge is armed.

andres-linero (approved):

PR #5147, opened 2026-09-07
chore(deps-dev): upgrade Vitest to 5

msaad00:
Summary
- Upgrades the UI test runner from Vitest 4.1.11 to 5.0.0.
- Uses the Vitest-specific Testing Library matcher entrypoint so the existing DOM assertions remain type-safe.
- Refreshes the owning product-proof manifest.

Verification
- Node 22.22.2 / npm 10.9.7
- npm ci --ignore-scripts
- npm run typecheck
- isolated overview reconciliation test: 10 passed
- full Vitest run: 1452 passed, 1 existing order-sensitive overview failure; isolated rerun passed 10/10
- npm run build
- npm run bundle:check
- npm run capture:product-proof
- make preflight

Notes
- The full-suite failure reproduces the existing order-sensitive overview test and passes when run in isolation; no product assertion failure was introduced by the Vitest upgrade.

### DIFF
--- a/docs/images/product-screenshots.json
+++ b/docs/images/product-screenshots.json
@@ -1,9 +1,9 @@
 {
   "release_version": "0.103.2",
-  "captured_at": "2026-09-07T06:58:34.010Z",
-  "source_commit": "1205746e5498aa1ad55d42431b5992f3dc863e79",
+  "captured_at": "2026-09-07T18:14:54.334Z",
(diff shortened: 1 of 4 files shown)


msaad00:
Consolidated into #5145. The combined UI dependency replacement now includes the Vitest 5 upgrade, the Testing Library matcher compatibility fix, all compatible UI dependency updates, current-main conflict repair, and refreshed product proof.

PR #5244, opened 2026-09-14
chore(deps): use one scheduler for routine updates

msaad00:
## Summary
- Remove the independent weekly uv.lock refresh schedule. Dependabot already groups routine updates into one weekly maintenance PR; a second scheduler created additional lockfile PRs.
- Retain workflow_dispatch for explicit full lockfile refreshes and update its PR wording.
- Guard the shared weekly group, one-open-PR limit, security-update groups, and manual-only lockfile refresh with a regression test.

## Verification
- `PYTHONPATH=src pytest -q tests/test_ci_workflow_contract.py` — 39 passed.
- `actionlint -shellcheck= .github/workflows/uv-lock-upgrade.yml`.
- `ruff check tests/test_ci_workflow_contract.py` and `ruff format --check tests/test_ci_workflow_contract.py`.
- `git diff --check`.

## Notes
Security-update configuration is unchanged. Already-open lockfile updates remain available for review. The next scheduled Dependabot run still needs operational verification of grouped PR creation.


msaad00:
Superseded by #5246. The combined replacement includes this update, repairs toolchain and MCP compatibility drift, and reconciles published registry counts. Local validation passes; CI and review continue on #5246.

